PR #12, opened 2020-03-22
Last part of iteration2

muuuydanny:


lib/turn.rb
```diff
@@ -0,0 +1,52 @@
...
+      if player1.deck.rank_of_card_at(2) > player2.deck.rank_of_card_at(2)
+        player1
+      else
+        player2
+      end
+    elsif @type == :mad
+      first_cards_are_same = player1.deck.rank_of_card_at(0) == player2.deck.rank_of_card_at(0)
+      second_cards_are_same = player1.deck.rank_of_card_at(2) == player2.deck.rank_of_card_at(2)
+      if first_cards_are_same && second_cards_are_same
+        player1.deck.remove_card
```
muuuydanny: hmm.... seems to be some duplicaysh (duplicated code, its a code smell all right) right here. Maybe use a loop ???

muuuydanny:
oops

PR #45, opened 2020-05-11
Game

IamNorma:
Added game class with methods:
- start
- play

### DIFF
--- a/test/game_test.rb
+++ b/test/game_test.rb
@@ -25,4 +25,23 @@ class GameTest < Minitest::Test
 
     assert_instance_of Game, game
   end
+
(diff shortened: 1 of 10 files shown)


test/game_test.rb
```diff
@@ -0,0 +1,47 @@
+require 'minitest/autorun'
+require 'minitest/pride'
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+require './lib/game'
+
+class GameTest < Minitest::Test
```
IamNorma, line 9: Probably best to write test to check if game can be won

PR #51, opened 2020-06-30
War_and_Peace first pull request

nicorithner:
@timomitchel


nicorithner:
@KathleenScriver

lib/card.rb
```diff
@@ -0,0 +1,20 @@
+class Card
+    attr_reader :suit, :value, :rank
```
KathleenScriver, line 2: What is this line accomplishing?

lib/card.rb
```diff
@@ -0,0 +1,20 @@
+class Card
+    attr_reader :suit, :value, :rank 
+    def initialize(suit, value, rank)
+        @suit = suit
+        @value = value 
+        @rank = rank
+    end
+
+    def suit
```
KathleenScriver, line 9: Looking at line 2, are these three methods needed?

lib/deck.rb
```diff
@@ -0,0 +1,28 @@
+class Deck
+  attr_reader :cards
+  def initialize(cards=[])
+  @cards = cards
+  end
+
+  def rank_of_card_at(index)
+    @cards[index].rank
```
KathleenScriver, line 8: Awesome

lib/deck.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def high_ranking_cards
+    @cards.map do |card|
+         card if card.rank >= 11   
+    end.compact
+  end
+  
+  def percent_high_ranking
+    ( ( high_ranking_cards.count * 100 ).to_f / @cards.count.to_f ).round(2)
```
KathleenScriver, line 18: With Ruby math, either the numerator OR denominator needs to be a float, but not both. :) So you can take out one of your `to_f` s

test/deck_test.rb
```diff
@@ -0,0 +1,87 @@
...
+        deck = Deck.new
+  
+        deck.cards << card1 = Card.new(:diamond, 'Queen', 12)
+        deck.cards << card2 = Card.new(:spade, '3', 3)
+        deck.cards << card3 = Card.new(:heart, 'Ace', 14)
+# binding.pry
+        assert_equal card1, deck.remove_card
+        assert_equal [card2, card3], deck.cards
+        assert_equal [card3], deck.high_ranking_cards
+        assert_equal 50.0, deck.percent_high_ranking
```
KathleenScriver, line 67: You don't need to test `high_ranking_cards` or `percent_high_ranking` here since those are already tested in their own tests. YOu'd just want to test `deck.cards` to make sure that `remove_card` actually did the desired action.

test/deck_test.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        assert_equal [card2, card3], deck.cards
+
+        card4 = Card.new(:club, '5', 5)
+
+# binding.pry
+        assert_equal [card2, card3, card4], deck.add_card(card4)
+        assert_equal [card2, card3, card4], deck.cards
+        assert_equal [card3], deck.high_ranking_cards
+        assert_equal 33.33, deck.percent_high_ranking
```
KathleenScriver, line 84: Same thing here.

lib/deck.rb
nicorithner: Oh yeah, that makes sense! thank you

PR #312, opened 2022-08-27
Amanda Ross

amikaross:


lib/turn.rb
```diff
@@ -0,0 +1,84 @@
...
+              :player2,
+              :spoils_of_war 
+
+  def initialize(player1, player2)
+    @player1 = player1
+    @player2 = player2
+    @spoils_of_war = []
+  end
+
+  def type
```
epintozzi, line 16: A few things are happening in this method that could be cleaned up. A nested `if` statement is sometimes necessary, however, if we can avoid it, that's even better. Nested `if` statements are prone to bugs and are hard to test. 

Another thing I notice that that there are two scenarios each that result in a type of `:war` and `:basic`. Ideally, our logic would have one path for every outcome (sometimes the ideal is not possible but is preferred when we can).

Something like this would accomplish the same thing but with fewer levels of logic:
```ruby
  def type
    if @player1.deck.rank_of_card_at(0) != @player2.deck.rank_of_card_at(0)
      :basic
    elsif (@player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)) && (@player1.deck.rank_of_card_at(2) == @player2.deck.rank_of_card_at(2)) && (player1.deck.cards.length >= 3 && player2.deck.cards.length >= 3)
      :mutually_assured_destruction
    elsif @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
      :war
    end
  end
```

I would also probably create some helper methods though to improve readability. One thing I noticed is that I played around with this logic a few different ways and your tests passed every time, even though I was missing the `cards.length >= 3` portion for a while. This tells me there could be more thorough testing. Tests should have failed when I forgot to include that piece of logic. 

A helper method would be something like:
```ruby
  def same_rank_at_zero?
    @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
  end
```

And then we can test that logic separately as well as call the `same_rank_at_zero?` method in our `if` statement instead of typing out that same comparison logic every time we need to check (currently happening on lines 18, 20, 26).

lib/turn.rb
```diff
@@ -0,0 +1,84 @@
...
+    else 
+      if player1.deck.rank_of_card_at(0) == player2.deck.rank_of_card_at(0)
+        :war 
+      else 
+        :basic 
+      end 
+    end
+  end
+
+  def winner
```
epintozzi, line 34: Similar to the `turn` method, I'm seeing a lot of nested logic here. Try playing around with helper methods to see if you can refactor this to be a little more digestible and easier to test.

lib/turn.rb
```diff
@@ -0,0 +1,84 @@
...
+      else
+        player2 
+      end
+    end
+  end
+
+  def pile_cards
+    if self.type == :basic
+      @spoils_of_war << @player1.deck.cards.shift
+      @spoils_of_war << @player2.deck.cards.shift
```
epintozzi, line 63: You can use the `remove_card` method from your `Deck` class here rather than rewriting the code (same for lines 67, 72).

lib/turn.rb
```diff
@@ -0,0 +1,84 @@
...
+         count += 1
+      end 
+    else 
+      3.times {@player1.deck.cards.delete_at(0)}
+      3.times {@player2.deck.cards.delete_at(0)}
+    end
+  end
+
+  def award_spoils(winner)
+    winner.deck.cards.concat(@spoils_of_war)
```
epintozzi, line 82: You should use your `add_card` method from `Deck` here rather than rewriting the code.

lib/game.rb
```diff
@@ -0,0 +1,44 @@
...
+require "./lib/player.rb"
+require "./lib/turn.rb"
+
+class Game 
+  def initialize(player1, player2)
+    @player1 = player1 
+    @player2 = player2
+  end
+
+  def start
```
epintozzi, line 12: There is A LOT happening in this method. I would break out the individual pieces of logic to their own methods so they can be better tested and easier to read.

spec/card_generator_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    it "creates readable cards" do 
+      filename = "cards.txt" 
+      cards = CardGenerator.new(filename).cards
+      expect(cards.length).to eq 52
+      hearts = cards[0..12]
+      clubs = cards[13..25]
+      diamonds = cards[26..38]
+      spades = cards[39..51]
+      values = ["2", "3", "4", "5", "6", "7", "8", "9", "10", "Jack", "Queen", "King", "Ace"]
+      hearts.each_with_index do |card, i|
```
epintozzi, line 21: This test is failing for me. Curious if it passes for you. For me, the cards are all off by one suit. Ex: The variable `hearts` are actually clubs.

epintozzi (commented):
Overall nice work! Your testing looks thorough. I would recommend starting to practice using helper methods so that you can test your code a bit more granularly, but that comes with practice. What you've submitted here is exactly what I expect to see at this point in the Turing program.

PR #313, opened 2022-08-28
Hilby - 2208, BE, Project 1

josephhilby:
Submitted for review.


lib/turn.rb
```diff
@@ -0,0 +1,76 @@
...
+      @player_one.deck.card_list.slice!(0, 3)
+
+      spoils_of_war << @player_two.deck.card_list[0..2]
+      @player_two.deck.card_list.slice!(0, 3)
+      @spoils_of_war.flatten!
+    elsif @turn_type == :mutually_assured_destruction
+      @player_one.deck.card_list.slice!(0, 3)
+      @player_two.deck.card_list.slice!(0, 3)
+    end
+  end
```
abdulredd, line 61: In this method I notice that you use the instance method `spoils_of_war` instead of the instance variable `@spoils_of_war`. This is the only instance variable that you do this with, any particular reasoning? This also happens on line 69 below.

lib/turn.rb
```diff
@@ -0,0 +1,76 @@
...
+  # were passed separated by one low card.
+  # Changed turn_spec.rb:125, 126, & 207 to match
+  def award_spoils
+    if @turn_type == :war || @turn_type == :basic
+    @turn_winner.deck.card_list << spoils_of_war.shuffle!
+    @spoils_of_war = []
+    @turn_winner.deck.card_list.flatten!
+    # @turn_winner = nil. No need. Each turn is an instance no need for nil.
+    # Still wrapping my head around this interaction.
+    end
```
abdulredd, line 74: Don't forget to indent your code properly for better readability. (lines 69, 70, 71)

Similar to what I mentioned earlier about `#flatten`. Adding the `@spoils_of_war` cards to the player's deck using `#concat` would have been a way to not have to flatten afterwards.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
+class Game
+  attr_reader :current_turn, :turn_count, :turn_return
+  def initialize(current_turn)
+    @current_turn = current_turn
+    @turn_count = 1
+    @turn_return = nil
```
abdulredd, line 6: Just reading this class from top to bottom, I'm not exactly sure what `@turn_return` is right now? What would have been a better, more descriptive name to use?

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  # Determine if a player has 0 cards in their deck
+  def the_end?
+    if @current_turn.player_one.has_lost? || @current_turn.player_two.has_lost?
+      true
+    end
+  end
```
abdulredd, line 14: So this predicate method, `the_end?`, only returns `true` or `nil`? What happened to `false`? Remember, in Ruby, it is customary that if a method name ends with a `?`, then it is assumed that it will return either `true` or `false`. We want to make sure to stick with conventions so other Ruby developers can make assumptions of our code and interface with it properly.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+        puts "Turn #{@turn_count}: *mutually assured destruction* 6 cards removed from play"
+        @turn_return = "Turn #{@turn_count}: *mutually assured destruction* 6 cards removed from play"
+      elsif @current_turn.turn_type == :loss
+        if @current_turn.player_one.deck.card_list.length > @current_turn.player_two.deck.card_list.length
+          puts "*~*~*~* #{@current_turn.player_one.name} has won the game! *~*~*~*"
+          break
+        elsif current_turn.player_one.deck.card_list.length < @current_turn.player_two.deck.card_list.length
+          puts "*~*~*~* #{@current_turn.player_two.name} has won the game! *~*~*~*"
+          break
+        end
```
abdulredd, line 41: Another instance where you do not require the `elsif` and can just use an `else` because the first `if` will either be `true` or it won't.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+
+      @current_turn.award_spoils
```
abdulredd, line 44: Great idea to only award spoils after the output is displayed to avoid any "turn" issues.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      @turn_count += 1
+
+      if the_end? == true
```
abdulredd, line 47: With the way that you've written the `the_end?` method, we know that the method will either return `true` or `nil` so we do not need to explicitly compare the method return value to `true`. `if the_end?` also works.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+        puts "*~*~*~* #{@current_turn.turn_winner.name} has won the game! *~*~*~*"
+        break
+      elsif @turn_count == 1_000_000
```
abdulredd, line 50: This is a great example of where an `elsif` is actually a great idea because we are checking an entirely different condition compared to the first `if` statement.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  # Loop to ensure correct user input
+  def start
+    puts 'Type "GO" to start the game!'
+    loop do
+      puts '-------------------------------------------------------------------'
+      user_input = gets.chomp.downcase
+      if user_input == 'go'
+        self.run
```
abdulredd, line 64: `self` is implied and therefore doesn't have to be mentioned. Just the word `run` by itself is the exact same thing.

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+        break
```
abdulredd, line 65: I'm wondering, under what circumstances would we ever reach this `break`?

lib/game.rb
```diff
@@ -0,0 +1,71 @@
...
+      else
+        puts 'Not a valid input. Please type "GO"'
```
abdulredd, line 67: Nitpicky:

```ruby
puts 'Not a valid input. Please type "GO" to start the game!'
```

lib/card_generator.rb
```diff
@@ -0,0 +1,47 @@
+require 'json'
+
+class CardGenerator
+  attr_reader :cards, :loaded_deck, :card_inputs
+  def initialize(cards = nil)
```
abdulredd, line 5: Your solution for iteration for doesn't quite follow the interaction pattern exactly.

The pattern looks like so:

```ruby
pry(main)> require './lib/card_generator'
#=> true

pry(main)> filename = "cards.txt"
#=> "cards.txt"

pry(main)> cards = CardGenerator.new(filename).cards
#=> [#<Card:0x007f9f1413cbe8...>, ... #<Card:0x007f9f1413c2b0...>]
```

I am supposed to be able to create a new instance of `CardGenerator` and pass it the filename that I want it to read and it should return an array of cards. The way your `CardGenerator` class is set up, it is expecting an array of cards passed in when you go to create an instance of `CardGenerator`, which is not expected behavior.

war_and_peace_runner.rb
```diff
@@ -0,0 +1,38 @@
+require 'pry'
+require './lib/card'
```
abdulredd, line 2: Make sure to get rid of any unnecessary `require`'s when turning in your submission. And keep in mind, since we are not "new-ing up" any `Card` objects inside this runner file, the `require './lib/card'` line can also be removed.

abdulredd (commented):
Overall I am really impressed with what you were able to put together here. It displays complete confidence in your programming abilities. I would just recommend you remove all your `require 'pry'` lines after you're done with them.

Your runner file is good, but I'm just left wondering why we are using one `Turn` for the entire game. One `Game` makes sense, but shouldn't each `Turn` be independent of the last? I also believe that some of this code could have been abstracted to the `Game` class to clean this file up a little more.

Lastly, I just want to mention how thoroughly impressed I am with your test and coverage. I looked through the majority of them and couldn't find anything out of place or feedback for improvement so great job there.

Keep up the great work!

PR #314, opened 2022-08-28
Ryan Loughry - War or Peace

thecoderanger5280:
I realize that some of my code that is commented out is more complex than it needs to be.


lib/deck.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def percent_high_ranking
+    high_cards = []
+    percent = 0.0
+    @cards.each do |card|
+      if(card.rank > 10)
+        high_cards << card
+      end
+    end
```
jamisonordway, line 28: consider calling `high_ranking_cards` here instead to make your code more DRY.

lib/turn.rb
```diff
@@ -0,0 +1,92 @@
...
+        :basic
+      # elsif((@player1.deck.cards.length < 3) || (@player2.deck.cards.length < 3))
+      #   :loss
+    elsif((@player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)) &&
+      (@player1.deck.rank_of_card_at(2) == @player2.deck.rank_of_card_at(2)))
+        :mutually_assured_destruction
+      elsif(@player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0))
+        :war
+      end
+    end
```
jamisonordway, line 25: Ideally, we would see `elsif` used instead of `else` with a nested `if`:
```
if (@player1.deck.cards.length < 3) || (@player2.deck.cards.length < 3)
  :sudden_death
elsif @player1.deck.rank_of_card_at(0) != @player2.deck.rank_of_card_at(0)
  :basic
elsif ((@player1.deck.rank_of_card_at(0)) == (@player2.deck.rank_of_card_at(0))) && ((@player1.deck.rank_of_card_at(2)) == (@player2.deck.rank_of_card_at(2)))
  :mutually_assured_destruction
elsif @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
  :war
end
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+        card6 = Card.new(:diamond, '8', 8)
+        card7 = Card.new(:heart, '3', 3)
+        card8 = Card.new(:diamond, '2', 2)
+        deck1 = Deck.new([card1, card2, card5, card8])
+        deck2 = Deck.new([card4, card3, card6, card7])
+        player1 = Player.new('Ryan', deck1)
+        player2 = Player.new('Ash', deck2)
+        turn = Turn.new(player1, player2)
+        new_deck = turn.player1.deck.cards.shuffle
+        require 'pry'; binding.pry
```
jamisonordway, line 130: don't forget to remove your `binding.pry`s when you're done with them.

lib/card_generator.rb
```diff
@@ -0,0 +1,38 @@
+require './lib/card'
+
+class Card_Generator
```
jamisonordway, line 3: This class should be called `CardGenerator`. While the name of this **file** is conventional, Class names use Pascal Case in Ruby

PR #315, opened 2022-08-28
Kristen Nestler

knestler:


lib/deck.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def rank_of_card_at(index)
+    @cards[index].rank
+  end
+
+  def high_ranking_cards
+    @cards.find_all do |card|
+    card.rank >= 11
+    end
```
jamisonordway, line 18: Great choice for an enumerable! Just keep an eye on your indentation:
```
@cards.find_all do |card|
  card.rank >= 11
end
```

lib/deck.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def percent_high_ranking
+    (high_ranking_cards.count.to_f / @cards.count*100).round(2)
```
jamisonordway, line 22: Nice job using the return value of `high_ranking_cards` here and keeping things DRY.

lib/player.rb
```diff
@@ -0,0 +1,13 @@
...
+
+attr_reader :name, :deck
+
+  def initialize(name, deck)
+    @name = name
+    @deck = deck
+  end
+
+  def has_lost?
+    deck.cards == []
```
jamisonordway, line 11: this is great! You could also do something like `deck.cards.empty?`

lib/turn.rb
```diff
@@ -0,0 +1,55 @@
...
+      @spoils_of_war << @player1.deck.cards.shift << @player2.deck.cards.shift
+    elsif type == :war
+      3.times do
+      @spoils_of_war << @player1.deck.cards.shift << @player2.deck.cards.shift
+      end
+    elsif type == :mutually_assured_destruction
+      3.times do
+      @player1.deck.cards.shift
+      @player2.deck.cards.shift
+      end
```
jamisonordway, line 48: (also applies to line 42) keep an eye on your indentation in these loops.
```
3.times do 
  @player1.deck.cards.shift
  @player2.deck.cards.shift
end
```
I do like the way you're using `times` loops in this method 👍🏻

lib/turn.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  def award_spoils(winner)
+    winner.deck.cards << @spoils_of_war58
```
jamisonordway, line 53: Is it possible that `@spoils_of_war58` is a typo?

PR #316, opened 2022-08-28
Matisse Mallette

MatisseMallette:


lib/deck.rb
```diff
@@ -0,0 +1,47 @@
+class Deck
+  # will need this later, could probably find a much better way to do this
+  attr_accessor :cards
+
+  def initialize(cards = [])
```
jamisonordway, line 5: great use of a default argument!

lib/deck.rb
```diff
@@ -0,0 +1,47 @@
...
+    @cards = cards
+  end
+
+  def rank_of_card_at(index)
+    @cards[index].rank
+  end
+
+  def high_ranking_cards
+    @cards.select { |card| card.rank >= 11 }
```
jamisonordway, line 14: Elegant use of the `select` enumerable on one line 🎉

lib/deck.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def percentage_high_ranking
+    self.high_ranking_cards.length.to_f / @cards.length.to_f
```
jamisonordway, line 18: Good job re-using the return value of the `high_ranking_cards` method here to keep your code DRY.

lib/player.rb
```diff
@@ -0,0 +1,17 @@
...
+  attr_reader :name, :deck
+
+  def initialize(name, deck)
+    @name = name
+    @deck = deck
+  end
+
+  def has_lost?
+    deck.cards.length == 0
+  end
```
jamisonordway, line 11: You could also do something like:
```
def has_lost?
  deck.cards.empty?
end
```

lib/turn.rb
```diff
@@ -0,0 +1,97 @@
...
+    @type = :basic
+    if @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
+      @type = :war
+      if @player1.deck.cards.length >= 3 && @player2.deck.cards.length >= 3
+        if @player1.deck.rank_of_card_at(2) == @player2.deck.rank_of_card_at(2)
+          @type = :mutually_assured_destruction
+        end
+      end
+    end
+    return @type
```
jamisonordway, line 27: Think about how you might be able to accomplish the same outcome without nesting `if` statements.

lib/turn.rb
```diff
@@ -0,0 +1,97 @@
...
+    # ALL OF THIS CAN BE DONE BETTER WITH SPLICE METHOD I THINK
+    type = self.type
+    if type == :basic
+      @spoils_of_war.add_card_to_bottom(@player1.deck.remove_card_from_top)
+      @spoils_of_war.add_card_to_bottom(@player2.deck.remove_card_from_top)
+    elsif type == :war
+      @spoils_of_war.add_card_to_bottom(@player1.deck.remove_card_from_top) unless @player1.deck.cards.empty?
+      @spoils_of_war.add_card_to_bottom(@player1.deck.remove_card_from_top) unless @player1.deck.cards.empty?
+      @spoils_of_war.add_card_to_bottom(@player1.deck.remove_card_from_top) unless @player1.deck.cards.empty?
+
```
jamisonordway, line 75: Consider using a `times` loop for this or something similar.

PR #317, opened 2022-08-28
Rich Kaht - War

Freeing3092:


lib/deck.rb
```diff
@@ -0,0 +1,38 @@
+class Deck
+  attr_reader :cards, :high_ranked_cards
+
+  def initialize(cards = [])
```
abdulredd, line 4: I see that you're adding a default argument here for `cards`, what was the thought process behind this? What stopgaps have been added to `Deck` to be triggered in the scenario where a new `Deck` instance doesn't receive an array of cards?

lib/deck.rb
```diff
@@ -0,0 +1,38 @@
...
+    @cards = cards
+    @high_ranked_cards = []
+  end
+  
+  def rank_of_card_at(index_position)
+    cards[index_position].rank
+  end
+  
+  def high_ranking_cards
```
abdulredd, line 13: I noticed that you added an instance variable called `@high_ranked_cards` to store your high ranking cards. With the way the method is written this isn't necessary.

```ruby
def high_ranking_cards
  high_ranked_cards = [] #<== Using a local variable
  cards.each do |card|
    if card.rank >= 11
      high_ranked_cards << card #<== Adding cards to that local variable
    end
  end
  return high_ranked_cards #<== Returning that local variable at the end
end
```

We can take it a step further and use an even better enumerable for the situation, `#select` or `#find_all`, they are aliases.

```ruby
def high_ranking_cards
  cards.select do |card|
    card.rank >= 11
  end
end
```

The `select/find_all` enumerable only returns the elements that meet the condition. (It does not modify the original array)

lib/deck.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+  
+  def remove_card
+    cards.shift
+  end
+  
+  def add_card(card)
+    cards.push(card)
+  end
+end
```
abdulredd, line 36: Last thing I want to mention is that if we are inside the class, we can use the instance variable `@cards` and reserve the instance method `cards` for calls outside of the class.

lib/deck.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+    end
+    return high_ranked_cards
+  end
+  
+  def percent_high_ranking
+    high_ranking_cards
+    percentage_high_ranked = high_ranked_cards.count / cards.count.to_f
+    (percentage_high_ranked * 100).round(2)
+  end
```
abdulredd, line 27: With the refactor I suggested earlier with `high_ranking_cards` this method could also be refactored like so:

```ruby
def percent_high_ranking
  percentage_high_ranked = high_ranking_cards.count / cards.count.to_f
  (percentage_high_ranked * 100).round(2)
end
```

lib/turn.rb
```diff
@@ -0,0 +1,121 @@
...
+  
+  # Return boolean value if either player has fewer than 3 cards left
+  
+  def fewer_than_3_cards?
+    if player1.deck.cards.count < 3 || player2.deck.cards.count < 3
+      true 
+    else
+      false
+    end
+  end
```
abdulredd, line 19: I really like the premise of this method and checking if a player even has enough cards to continue, just want to suggest how you can make this method more succinct. Since we know that `player1.deck.cards.count < 3 || player2.deck.cards.count < 3` will either return `true` or `false`, we can use that to our advantage and refactor this method to look like so:

```ruby
def fewer_than_3_cards?
  player1.deck.cards.count < 3 || player2.deck.cards.count < 3
end
```

Once this method runs, the result will either be `true` or it will be `false`.

lib/turn.rb
```diff
@@ -0,0 +1,121 @@
...
+    if player1.deck.cards[0].rank == player2.deck.cards[0].rank && fewer_than_3_cards?
+      return :sudden_death
+    elsif player1.deck.cards[0].rank == player2.deck.cards[0].rank && player1.deck.cards[2].rank == player2.deck.cards[2].rank
+      return :mutually_assured_destruction
+    elsif player1.deck.cards[0].rank == player2.deck.cards[0].rank
+      return :war
+    else
+      return :basic
+    end
+  end
```
abdulredd, line 34: I like the sequence of these conditionals, starting with checking if each player has enough cards to continue, and then moving onto the `:mutually_assured_destruction` scenario, and then `:war`, and finally `:basic`.

lib/turn.rb
```diff
@@ -0,0 +1,121 @@
...
+        @spoils_of_war << @player1.deck.cards.shift
+        @spoils_of_war << @player2.deck.cards.shift
+      end
+    elsif type == :mutually_assured_destruction
+      3.times do
+        @player1.deck.cards.shift
+        @player2.deck.cards.shift
+      end
+    end
+  end
```
abdulredd, line 86: Throughout this method we used `#deck.cards.shift`. Don't we have a method on `Deck` that does this shifting thing already?

```ruby
def remove_card
  cards.shift
end
```

lib/turn.rb
```diff
@@ -0,0 +1,121 @@
...
+  
+  # Remove cards from spoils of war and add to winning player deck
+  
+  def award_spoils(winner)
+    @spoils_of_war.count.times do |card|
```
abdulredd, line 91: I wonder why we are opting for the `#times` method here? Would `#each` have been a better choice?

```ruby
def award_spoils(winner)
  @spoils_of_war.each do |card|
    @turn_winner.deck.cards << card
  end
  @spoils_of_war.clear
end
```

Also, the method is accepting an argument, `winner` but that argument isn't being used at all. What advantages are we gaining with the `@turn_winner` instance variable?

lib/game.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+  
+  # Set turn count to 0 and play out a game until the earlier of either player losing, or 1,000,000 turns
+  
+  def play
+    turn_count = 0
+    
+    until player1.has_lost? || player2.has_lost? do
+      turn_count += 1
+      turn = Turn.new(player1, player2)
```
abdulredd, line 30: I really really like that for each turn you are creating a new `Turn` object. This is exactly how the game is supposed to flow programmatically.

lib/game.rb
```diff
@@ -0,0 +1,57 @@
...
+      elsif turn.type == :basic
+        puts "Turn #{turn_count}: #{turn.winner.name} won 2 cards!\n"
+      elsif turn.type == :war
+        puts "Turn #{turn_count}: WAR - #{turn.winner.name} won 6 cards\n"
+      elsif turn.type == :mutually_assured_destruction
+        puts "Turn #{turn_count}: *mutually assured destruction* 6 cards removed from play\n"
+      end
+      
+      turn.pile_cards
+      turn.award_spoils(@turn_winner)
```
abdulredd, line 44: Piling the cards and awarding the spoils after the output to the CLI is a very smart move!

lib/game.rb
```diff
@@ -0,0 +1,57 @@
...
+      
+      if player1.has_lost? || player2.has_lost?
+        puts "*~*~*~* #{player1.deck.cards[0].nil? ? player2.name : player1.name} has won the game! *~*~*~*"
```
abdulredd, line 47: Ternary operator FTW! They're not appropriate everywhere but in the few places they are, they look pretty sweet.

lib/card_generator.rb
```diff
@@ -0,0 +1,23 @@
+require './lib/card'
+
+class CardGenerator
+  attr_reader :cards, :filename
+  def initialize(filename)
+    @filename = filename
+    @cards = cards
```
abdulredd, line 7: Where is the variable `cards` coming from? Would it have been better to start the instance variable `@cards` off as an empty array and shovel (`<<`) cards into it as they were being created in your `sort_file_into_deck` method? This would eliminate the need for the `card_collector` variable as well.

lib/card_generator.rb
```diff
@@ -0,0 +1,23 @@
...
+    @cards = sort_file_into_deck()
```
abdulredd, line 8: Woah, 2 instance variables with the same exact name?

spec/deck_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    card2 = Card.new(:heart, '3', 3)
+    card3 = Card.new(:diamond, 'Queen', 12)
+    cards = [card1, card2, card3]
+    deck = Deck.new(cards)
+    
+    card4 = Card.new(:club, '7', 7)
+    deck.add_card(card4)
+    
+    expect(cards[3]).to eq(card4)
+  end
```
abdulredd, line 69: This test should make an assertion for before `card4` is added to the `deck` and after `card4` is added to the `deck`.

```ruby
it 'has a method to add one card to the bottom of the deck' do
  card1 = Card.new(:spade, 'Ace', 14)
  card2 = Card.new(:heart, '3', 3)
  card3 = Card.new(:diamond, 'Queen', 12)
  cards = [card1, card2, card3]
  deck = Deck.new(cards)

  expect(deck.cards).to eq(cards)

  card4 = Card.new(:club, '7', 7)
  deck.add_card(card4)

  expect(deck.cards).to eq([card1, card2, card3, card4])
end
```

While this test and the one you wrote essentially perform the same task, these assertions are a little better. `expect(cards[3]).to eq(card4)` isn't the best assertion. Our assertions should really be testing method calls.

spec/player_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+RSpec.describe Player do
+  before(:each) do
+    card1 = Card.new(:spade, 'Ace', 14)
+    card2 = Card.new(:heart, '3', 3)
+    card3 = Card.new(:diamond, '8', 8)
+    cards = [card1, card2, card3]
+    @deck = Deck.new(cards)
+    @player = Player.new('John', @deck)
+  end
```
abdulredd, line 14: Great job including a `before(:each)` block here to clean up your code and make it more DRY.

abdulredd (commented):
Overall, great job on your project, I was able to watch a few games get played and the game never broke. There was a situation where after a couple thousand turns there would just be a ping-ponging until we eventually got to 1,000,000 turns and the game would just end in a draw.

Kudos on creating your own custom tests and not relying on the interaction pattern. I looked through most of it and apart from the comments I made, it was great set up and assertions. I couldn't find a `card_spec.rb` file in your project though.

Great jog, keep up the great work!

PR #318, opened 2022-08-28
Sean Culliton

smculliton:


lib/player.rb
```diff
@@ -0,0 +1,16 @@
+require 'pry'
+require_relative './deck'
```
abdulredd, line 2: `require`'s aren't necessary or need to be removed.

lib/player.rb
```diff
@@ -0,0 +1,16 @@
...
+class Player
+    attr_reader :deck, :name
+
+    def initialize(name, deck)
+        @name = name
+        @deck = deck 
+    end 
+
+    def has_lost?
+        return true if @deck.cards.none?
```
abdulredd, line 13: I love the use of explicit and implicit `return`s here. One thing to keep in mind is that `none?` is a predicate/boolean method and therefore we know that it will return either `true` or `false` after it is called, we can use that to our advantage and rewrite the code like so:

```ruby
def has_lost?
  @deck.cards.none?
end
```

The result will either be `true` or `false`.

lib/turn.rb
```diff
@@ -0,0 +1,61 @@
+require 'pry'
+require_relative './card'
+require_relative './deck'
+require_relative './player'
```
abdulredd, line 4: Inside the `Turn` class we never create a `Card`, `Deck`, or `Player` object therefore no need for the `require`'s.

lib/turn.rb
```diff
@@ -0,0 +1,61 @@
...
+    def initialize(player1, player2)
+        @player1 = player1
+        @player2 = player2
+        @spoils_of_war = []
+    end
+
+    def type
+        if @player1.deck.rank_of_card_at(0) != @player2.deck.rank_of_card_at(0)
+            return :basic
+        elsif @player1.deck.cards[2] == nil || @player2.deck.cards[2] == nil
```
abdulredd, line 19: I like this extra condition that checks whether a player even has enough cards in their deck and then performing some alternative action based on this condition.

lib/turn.rb
```diff
@@ -0,0 +1,61 @@
...
+            return :redo
+        elsif @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0) && @player1.deck.rank_of_card_at(2) == @player2.deck.rank_of_card_at(2)
+            return :mutually_assured_destruction 
+        else @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
+            return :war
+        end 
+    end 
+
+    #I added a redo turn type in this because I was receiving error messages when one player had < 3 cards and a war turn got called.
```
abdulredd, line 28: Makes sense to me. That bug was bugging the hell out of me.

lib/turn.rb
```diff
@@ -0,0 +1,61 @@
...
+        if type == :redo
+            return [@player1, @player2].each { |player| player.deck.top_card_to_bottom }
+        elsif type == :basic
+            return [@player1, @player2].each { |player| @spoils_of_war << player.deck.remove_card } 
+        elsif type == :war
+            return [@player1, @player2].each { |player| 3.times { @spoils_of_war << player.deck.remove_card } } 
+        else 
+            return [@player1, @player2].each { |player| 3.times { player.deck.remove_card } }
+        end 
+    end
```
abdulredd, line 54: Woah, I love how clean this looks!! Way to cut down the repetition bit looping over each player.

lib/turn.rb
```diff
@@ -0,0 +1,61 @@
...
+        elsif type == :war
+            return [@player1, @player2].each { |player| 3.times { @spoils_of_war << player.deck.remove_card } } 
+        else 
+            return [@player1, @player2].each { |player| 3.times { player.deck.remove_card } }
+        end 
+    end 
+
+    def award_spoils(winner)
+        return if winner == 'No winner' || winner == 'redo'
+        @spoils_of_war.shuffle.each { |spoil| winner.deck.cards << spoil }
```
abdulredd, line 58: Why was it necessary to shuffle the deck? I know you mentioned it helped to avoid an infinite game, but I'm just wondering how an infinite game would be possible if one person would eventually have more cards than the other person.

lib/card_generator.rb
```diff
@@ -0,0 +1,21 @@
+require 'pry'
+require './lib/card'
```
abdulredd, line 2: This `require` IS required because we are creating **new** `Card` objects inside of this class. The `pry` `require` can go though.

lib/card_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class CardGenerator
+    attr_reader :file, :cards
```
abdulredd, line 5: In some classes, you include a space after your `attr_reader` line, and in some, you don't. I'd pick a style and stick with it. It is convention to add a blank line after the `attr_reader` line.

I'm not sure we need an `attr_reader` for `@file` since it doesn't say in the interaction pattern that this attribute will ever be called from outside this class.

lib/card_generator.rb
```diff
@@ -0,0 +1,21 @@
...
+    def generate_deck
+        cards = []
+        @file.each do |line|
+             card = line.split(',').map { |item| item.strip }
+             card[2] = card[2].to_i
+             card[1] = card[1].downcase.to_sym
+             cards << Card.new(card[1],card[0],card[2])
+        end
+        cards
+    end
```
abdulredd, line 20: This method is very clean and easy to read. Great job!

spec/deck_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    end 
+
+    describe '#remove_card' do 
+        it 'removes and returns first card' do 
+            @deck.remove_card
+            expect(@deck.cards).to eq([@card2, @card3])
+        end 
+
+        it 'returns first card' do 
+            card = @decl.remove_card
```
abdulredd, line 58: `@deck`. You want to make sure to run your tests one final time before submitting your work to make sure there aren't things you missed still lurking.

spec/card_spec.rb
```diff
@@ -1,18 +1,20 @@
 require 'rspec'
-require './lib/card'
+require '../lib/card'
 
 RSpec.describe Card do
-  it "exists" do
-    card = Card.new(:diamond, 'Queen', 12)
+  describe "#initialize" do
```
abdulredd, line 5: I like how you refactored this spec to use a `describe` block for organization.

spec/deck_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'rspec'
+require '../lib/deck'
+require '../lib/card'
+
+RSpec.describe Deck do 
+    before(:each) do
```
abdulredd, line 6: Great job using a `before` block to reduce repetition and make your code DRY.

spec/deck_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'rspec'
+require '../lib/deck'
+require '../lib/card'
+
+RSpec.describe Deck do
```
abdulredd, line 5: Super clean, having each `describe` represent a method that you are testing.

war_or_peace_runner.rb
```diff
@@ -0,0 +1,69 @@
+require 'pry'
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+require './lib/card_generator'
+
+class Game
```
abdulredd, line 8: This class should be in a file called `game.rb` and the only code that should be in this runner are what is necessary to start the game which I imagine would look something like this:

```ruby
require './lib/game'
Game.new
```

A runner file is just a jumping-off point, not a place to define classes and implementation code.

Moving beyond that though, I like the way that this class is set up, all your variables and methods are named appropriately and help to explain what responsibility each one has. Very easy on the eyes and easy to get through.

war_or_peace_runner_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rspec'
+require './lib/card'
+require './war_or_peace_runner'
+
+RSpec.describe Game do
```
abdulredd, line 5: This file would be `game_spec.rb` after refactoring the `Game` class into it's own file.

Runner files do not need testing.

abdulredd (commented):
Overall, this was a great project and I really liked your tests. You tested various scenarios, and included expectations for before and after a method was run. Keep up the great work!

PR #320, opened 2022-08-28
Ashley Turner

ashuhleyt:
I left some comments in the game.rb, but I am meeting with my mentor this week to finish i3. I struggled to piece together how to get the game to actually run. I would love feedback on the other code I wrote. Thank you!

### DIFF
--- a/lib/card.rb
+++ b/lib/card.rb
@@ -1,0 +1,11 @@
+class Card 
+  attr_reader :suit, 
+              :value, 
+              :rank
+              
+  def initialize(suit, value, rank)
+    @suit = suit 
+    @value = value
+    @rank = rank 
+  end
+end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -31,9 +31,3 @@ end
 
 
 
- # last_digit = (high_ranking_cards.length / @cards.length.to_f * 100).to_s.slice(-1).to_i
-    # if last_digit >= 5
-    #   (high_ranking_cards.length / @cards.length.to_f * 100).ceil(2)
-    # else 
-    #   (high_ranking_cards.length / @cards.length.to_f * 100).floor(2)
-    # end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -16,15 +16,24 @@ class Deck
   end
 
   def percent_high_ranking
-    last_digit = (high_ranking_cards.length / @cards.length.to_f * 100).to_s.slice(-1).to_i
-    if last_digit >= 5
-      (high_ranking_cards.length / @cards.length.to_f * 100).ceil(2)
-    else 
-      (high_ranking_cards.length / @cards.length.to_f * 100).floor(2)
-    end
+    (high_ranking_cards.length.to_f / @cards.length.to_f * 100).round(2)
   end
 
   def remove_card
     @cards.shift 
   end
+
+  def add_card(card)
+    @cards.push(card)
+  end
 end
+
+
+
+
+ # last_digit = (high_ranking_cards.length / @cards.length.to_f * 100).to_s.slice(-1).to_i
+    # if last_digit >= 5
+    #   (high_ranking_cards.length / @cards.length.to_f * 100).ceil(2)
+    # else 
+    #   (high_ranking_cards.length / @cards.length.to_f * 100).floor(2)
+    # end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -16,6 +16,15 @@ class Deck
   end
 
   def percent_high_ranking
-    
+    last_digit = (high_ranking_cards.length / @cards.length.to_f * 100).to_s.slice(-1).to_i
+    if last_digit >= 5
+      (high_ranking_cards.length / @cards.length.to_f * 100).ceil(2)
+    else 
+      (high_ranking_cards.length / @cards.length.to_f * 100).floor(2)
+    end
+  end
+
+  def remove_card
+    @cards.shift 
   end
 end
--- a/lib/deck.rb
+++ b/lib/deck.rb
@@ -1,0 +1,21 @@
+class Deck 
+  attr_reader :cards
+
+  def initialize(cards) 
+    @cards = cards 
+  end
+
+  def rank_of_card_at(index)
+    @cards[index].rank
+  end
+
+  def high_ranking_cards
+    @cards.select do |card|
+      card.rank >= 11
+    end
+  end
+
+  def percent_high_ranking
+    
+  end
+end
--- a/lib/game.rb
+++ b/lib/game.rb
@@ -1,6 +1,91 @@
-# class Game 
+require 'rspec'
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+require './lib/game'
+
+class Game 
+  attr_reader :deck, :player1, :player2, :turn
+
+  def initialize 
+    @deck = []
+    @player1 = Player.new("Megan", Deck.new(deck[0..25]))
+    @player2 = Player.new("Aurora", Deck.new(deck[26..51]))
+    @turn = Turn.new(@player1, @player2)
+  end
+
+  def create_deck
+
+    @deck << @card1 = Card.new(:spade, '2', 2)
+    @deck << @card2 = Card.new(:spade, '3', 3)
+    @deck << @card3 = Card.new(:spade, '4', 4)
+    @deck << @card4 = Card.new(:spade, '5', 5)
+    @deck << @card5 = Card.new(:spade, '6', 6)
+    @deck << @card6 = Card.new(:spade, '7', 7)
+    @deck << @card7 = Card.new(:spade, '8', 8)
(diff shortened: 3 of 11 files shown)


lib/card.rb
```diff
@@ -0,0 +1,11 @@
...
+  attr_reader :suit, 
+              :value, 
+              :rank
+              
+  def initialize(suit, value, rank)
+    @suit = suit 
+    @value = value
+    @rank = rank 
+  end
+end
```
abdulredd, line 11: This class is perfect!

lib/deck.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def remove_card
+    @cards.shift 
+  end
+
+  def add_card(card)
+    @cards.push(card)
+  end
+end
```
abdulredd, line 29: Another perfect class! Great job using the instance variable `@cards` in your methods and keeping your code DRY by reusing the previously defined `high_ranking_cards` in your `percent_high_ranking` method. Also great choice in using the `select` enumerable in `high_ranking_cards`.

lib/player.rb
```diff
@@ -0,0 +1,13 @@
...
+  
+  def initialize(name, deck) 
+    @name = name 
+    @deck = deck 
+  end
+
+  def has_lost? 
+    @deck.cards.count == 0
+  end
+end
```
abdulredd, line 13: Another great class. Very clean.

lib/turn.rb
```diff
@@ -0,0 +1,73 @@
...
+      @spoils_of_war << @player2.deck.cards.shift()
+    end
+  end
+
+  def award_spoils(winner)
+    @spoils_of_war.each do |card|
+      winner.deck.cards << card
+    end
+    @spoils_of_war = []
+  end
```
abdulredd, line 60: This method is wonderfully written and works as intended if the game's are of a `:basic` or `:war` type. If the game is `:mutually_assured_destruction` the game breaks here. This is because, if you remember, when a game is `:mutually_assured_destruction` we get `"No Winner"` returned from the `winner` method. And therefore when we call `award_spoils` and pass it `winner`, it will have the value `"No Winner"` string and you cannot call `.deck` on a String.

One way to avoid this behavior is to add a conditional statement to this method so that way it only runs if an actual player object is stored in `winner`.

```ruby
def award_spoils(winner)
  if winner != "No Winner" #<== If winner is not equal to "No Winner"
    @spoils_of_war.each do |card| #<== then do this
      winner.deck.cards << card
    end
  end
  @spoils_of_war = []
end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+
+class Game 
+  attr_reader :deck, :player1, :player2, :turn
+
+  def initialize 
+    @deck = []
+    @player1 = Player.new("Megan", Deck.new(deck[0..25]))
+    @player2 = Player.new("Aurora", Deck.new(deck[26..51]))
+    @turn = Turn.new(@player1, @player2)
+  end
```
abdulredd, line 16: I'm going to try and get you unstuck for iteration 3.

Ok so when we call `Game.new` in your runner file, we know that the `initialize` method in your `Game` class is what gets run. In your initialize we set up a `@deck` instance variable to be an empty array (`[]`) and then on the next 2 lines we create two `Player` objects. When you're passing those `Player` objects `deck[0..25]` and `deck[0..25]` let's try and think about what we're actually passing.

What is the value of `deck` currently? It's empty... remember on line 12 we set it to be empty, so it still is empty. So when we are doing `Deck.new(deck[0..25])` or `Deck.new(deck[26..51])` we are passing 2 `Deck` objects with empty cards.

I see that you have a `create_deck` method defined that actually creates the deck of cards and shuffles it but it never gets called. So I would refactor your `initialize` method to look like this:

```ruby
def initialize 
  @deck = []
  create_deck # <== Notice that I add the call to the method that actually creates the deck of cards
  @player1 = Player.new("Megan", Deck.new(deck[0..25])) # <== Now deck[0..25] actually has cards in it
  @player2 = Player.new("Aurora", Deck.new(deck[26..51])) # <== deck[26..51] as well
  # @turn = Turn.new(@player1, @player2)
end
```

You'll notice that I commented out the `@turn` that is because I want to create the turn in the `start` method. Since I know I will need a loop, I want to create a new `turn` object for each.. well -- turn.

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+    @deck << @card52 = Card.new(:club, 'Ace', 14)
+    @deck.shuffle
+  end
+
+  def start 
+    p "Welcome to War! (or Peace) This game will be played with 52 cards."
+    p "The players today are Megan and Aurora."
+    p "Type 'GO' to start the game!"
+    p '------------------------------------------------------------------'
+  end
```
abdulredd, line 80: Ok, so your `start` method starts out great by outputting the message to the screen to greet the user. What you want to do now is get user input from them because we asked them for it. So here's what the start of your method should be refactored to look like:

```ruby
def start
  p "Welcome to War! (or Peace) This game will be played with 52 cards."
  p "The players today are Megan and Aurora."
  p "Type 'GO' to start the game!"
  p '------------------------------------------------------------------'

  user_input = gets.chomp.upcase # <== Here we ask user for input, we hope that they enter "GO" or some variation of it ("go", "Go", "gO"). It doesn't matter what case they use because we will upcase it ourselves

  until user_input == "GO" # <== We check if it is "GO" (after we upcase it)
    p "I'm sorry that input was invalid. Type 'GO' to start the game!" # <== If it's not "GO" we will print this msg
    user_input = gets.chomp.upcase # <== And get input again and again and again until they comply
  end

  # After the user enters the correct thing we will start the game...
end
```

lib/game.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  start = gets.chomp
+
+
+  # I am struggling with how to execute actual game play. have reached
+  #out to my mentor for additional support in understanding how to 
+  #incorporate everything to get the actual game to run. 
+  #
```
abdulredd, line 88: Ok so now that we have hopefully got correct input from the user it is time to run the game...

```ruby
def start
  # I'm just hiding the previous code in this method for space, it is still here...

  turn_num = 1

  until @player1.has_lost? || @player2.has_lost? || turn_num == 1_000_000 # <== We want this loop to keep going until one of the player loses or we reach 1000000 turns
    turn = Turn.new(@player1, @player2) # <== For each iteration of the loop we want to create a new turn, passing in the same players except they will have less and less cards each time after the spoils have been awarded
    
    # Next we want to check the `turn.type` and print out the correct phrase by that type
    if turn.type == :basic
      p "Turn #{turn_num}: Whatever the phrase for a basic turn was..."
    elsif turn.type == :war
      p "Turn #{turn_num}: Whatever the phrase for a war turn was..."
    elsif turn.type == :mutually_assured_destruction
      p "Turn #{turn_num}: Whatever the phrase for a MAD turn was..."
    end

    # Afterwhich we will pile the cards, award the spoils of war cards to the winner, increase the turn count by 1 and begin again
    turn.pile_cards
    turn.award_spoils(turn.winner)
    turn_num += 1
  end

  # Once someone has lost, we need to figure out who that player was and print the appropriate output
  if @player1.has_lost?
    p "#{@player2.name} wins!"
  else
    p "#{@player1.name} wins!"
  end

  if turn_num == 1_000_000
    p "-----DRAW-----"
  end
end
```

Now I'm not sure if the code will work as I've written here exactly but this is just about the gist of it.

spec/player_spec.rb
```diff
@@ -0,0 +1,42 @@
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+
+RSpec.describe do
```
abdulredd, line 5:
```ruby
RSpec.describe Player do 
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,171 @@
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+
+RSpec.describe do
```
abdulredd, line 6:
```ruby
RSpec.describe Turn do 
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+    winner = @turn.winner
+
+    @turn.pile_cards
+    @turn.award_spoils(winner)
+
+    expect(@player1.deck.cards).to eq([@card8])
+    expect(@player2.deck.cards).to eq([@card7])
+    end
+  end
+end
```
abdulredd, line 171: This `Turn` class spec is great but there are quite a few tests still missing.

abdulredd (commented):
All of your classes were crazy clean and the code was super easy and enjoyable to get through. It almost read like English. That shows a definite command in your ability to write nice Ruby code.

I did try an add some feedback about fleshing out the rest of your `Game` class, I hope you find it useful.

Your tests were also great, but your `Turn` spec was missing quite a few tests still. At the end of the day, you should feel proud about what you were able to accomplish here, I know I am.

Keep up the great work Ashley!

PR #321, opened 2022-08-28
Bryan Keener

bkeener7:


lib/player.rb
```diff
@@ -0,0 +1,13 @@
...
+    attr_reader :name, :deck
+    def initialize(name, deck)
+        @name = name
+        @deck = deck
+    end
+
+    def has_lost?
+        return true if deck.cards.length <= 0
+        false
+    end
```
abdulredd, line 11: Love the combination of explicit and implicit returns being leveraged in this method, nicely done. I do want to point out one tip, however. Since we know that `deck.cards.length <= 0` will either return `true` or `false`, we can use that to our advantage and refactor the method to look something like this:

```ruby
def has_lost?
  deck.cards.length <= 0
end
```

Once this method evaluates, it's either going to be `true` or `false` and then return it.

The last thing I want to mention is to remember that since we are already inside the class, we can use the instance variables. I know if you check online there's a back and forth as to the preferred way, but Turing pushes using the instance variable over the instance method if you are inside the class, at least in Mod 1.

```ruby
def has_lost?
  @deck.cards.length <= 0
end
```

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
+class Turn
+    attr_reader :player1, :player2, :spoils_of_war
+    def initialize(player1, player2)
+        @player1 = player1
+        @player2 = player2
+        @spoils_of_war = []
+        @turn_type = nil
```
abdulredd, line 7: Saving the turn type into an instance variable is a very smart move and that way you only have to call the `type` method once per turn. Great idea!

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    def type
+        if @player1.deck.rank_of_card_at(0) != @player2.deck.rank_of_card_at(0)
+            @turn_type = :basic
+        elsif @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0) && @player1.deck.cards.length < 3 || @player2.deck.cards.length < 3
```
abdulredd, line 13: I like that you include another condition to check if each player even has enough cards to continue before determining if it is a `:mutually_assured_destruction` or `:war` situation. I know this was giving a lot of people fits at the end when running the game.

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+            @player2
+        end
+    end
+
+    def pile_cards
+        if @turn_type == :mutually_assured_destruction
+            players = [@player1, @player2]
+            3.times do
+                players.each do |card|
+                    card.deck.cards.shift
```
abdulredd, line 42: I vaguely remember there being a method on the `Deck` class that we defined that already does this `cards.shift` thing... 🤔 

```ruby
def remove_card
  cards.shift
end
```

Also, I think the block variable `|cards|` is not very descriptive of what it actually is. If we are iterating over the array of `players` defined on line 39, wouldn't each iteration be a `|player|` then?

Using `3.times` was a great way to avoid repetition!

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+                end
+            end        
+        elsif @turn_type == :war
+            players = [@player1, @player2]
+            3.times do
+                players.each do |card|
+                    @spoils_of_war << card.deck.cards.shift
```
abdulredd, line 49: `Deck#remove_card`

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+            players = [@player1, @player2]
+            3.times do
+                players.each do |card|
+                    card.deck.cards.shift
+                end
+            end        
+        elsif @turn_type == :war
+            players = [@player1, @player2]
+            3.times do
+                players.each do |card|
```
abdulredd, line 48: `|player|`

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+                    @spoils_of_war << card.deck.cards.shift
+                end
+            end
+        elsif @turn_type == :basic
+            players = [@player1, @player2]
+            players.each do |card|
```
abdulredd, line 54: `|player|`

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+                @spoils_of_war << card.deck.cards.shift
```
abdulredd, line 55: `Deck#remove_card`

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+            end
+        end
+    end
```
abdulredd, line 58: In this method, `pile_cards`, it looks like each `if' condition uses `players = [@player1, @player2]`. Would it have been better to define this on line 37.5 so that you don't have to repeat it inside every condition?

lib/turn.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    def award_spoils_of_war(winner)
+        if winner == "No Winner"
+            @spoils_of_war = []
+        else
+            (winner.deck.cards << @spoils_of_war.shuffle).flatten!
+            @spoils_of_war = []
+        end
+    end
```
abdulredd, line 67: In this method, both conditions seem to clear out the `@spoils_of_war` array. To avoid repetition, I think this method could be refactored to look something like this:

```ruby
def award_spoils_of_war(winner)
  if winner != "No Winner"
    (winner.deck.cards << @spoils_of_war.shuffle).flatten! #<== This will only run if there is a winner
  end
  @spoils_of_war = [] #<== This will always run, whether there is a winner or not
end
```

With all that being said, I am wondering why the `@spoils_of_war` needs to be emptied out anyway since it looks like the way you have the game running in your `Game#start` method, a new `turn` object is created for each new turn, so the clearing happens automatically because every new turn starts with an empty `@spoils_of_war`.

lib/game.rb
```diff
@@ -0,0 +1,131 @@
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+
+class Game
+    def initialize
```
abdulredd, line 7: If your `initialize` method is going to be empty, you do not need to define it at all. A class will work without it being explicitly defined.

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+
+    def welcome_message 
+        puts "Welcome to War! (or Peace) This game will be played with 52 cards."
+        puts "The players today are Megan and Aurora."
+        puts "Type 'GO' to start the game!"
+        puts "------------------------------------------------------------------"
+        user_input = gets.chomp.upcase
+
+        if user_input == "GO"
```
abdulredd, line 17: You have a, "you only get one re-do"-type game lol. You either enter the correct input on the second attempt or you don't get to play. That's fine, but I found that interesting because more often than not, you see users being given multiple attempts to enter incorrect input.

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+        player1 = Player.new("Megan", deck1)
+        player2 = Player.new("Aurora", deck2)
+    
+        26.times do
+            player1.deck.cards << full_deck.shift
+        end
+    
+        26.times do
+            player2.deck.cards << full_deck.shift
+        end
```
abdulredd, line 98: I think there might have been an easier way to disperse the cards amongst the players than creating 2 loops that each run 26 times. How could we have used array slicing to make this logic better?

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+                        turn.pile_cards
+                        p "Turn #{turn_number.to_s}: *mutually assured destruction* 6 cards removed from play"
+                    elsif current_turn == :war                    
+                        turn.pile_cards
+                        p "Turn #{turn_number.to_s}: WAR - #{winner.name} won #{turn.spoils_of_war.length} cards"
+                    elsif current_turn == :basic
+                        turn.pile_cards
+                        p "Turn #{turn_number.to_s}: #{winner.name} won #{turn.spoils_of_war.length} cards"
+                    end          
+                turn.award_spoils_of_war(winner)
```
abdulredd, line 128: Great job deciding to award the spoils after all the output has been printed so that you don't get any unexpected output.

lib/game.rb
```diff
@@ -0,0 +1,131 @@
...
+            end                   
+    end
```
abdulredd, line 130: I don't see anything amiss with this `start` method. I will mention a few things though:
* This method is super long and should be broken down into smaller helper methods
* The indentation in this method is a little off

spec/deck_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+RSpec.describe Deck do
+    it 'has cards' do
+        card1 = Card.new(:diamond, 'Queen', 12)
+        card2 = Card.new(:spade, '3', 3)
+        card3 = Card.new(:heart, 'Ace', 14)        
+
+        expect(card1).to be_an_instance_of(Card)
+        expect(card2).to be_an_instance_of(Card)
+        expect(card3).to be_an_instance_of(Card)
+    end
```
abdulredd, line 14: This test looks like it would be better suited in a `cards_spec.rb` file.

spec/deck_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    it 'has a certain percentage of high ranking cards' do
+        card1 = Card.new(:diamond, 'Queen', 12)
+        card2 = Card.new(:spade, '3', 3)
+        card3 = Card.new(:heart, 'Ace', 14)
+        deck = Deck.new([card1, card2, card3])
+
+        # expect 66.6% of cards to be high ranking (card1 & card3)
+        expect(deck.percent_high_ranking).to eq(2/3.to_f * 100)
```
abdulredd, line 52: We always want our `eq` to be an actual value we are expecting and not a calculation like it is here.

```ruby
expect(deck.percent_high_ranking).to eq(66.6)
```

spec/player_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+RSpec.describe Player do
+    it 'has cards and a deck' do
+        card1 = Card.new(:diamond, 'Queen', 12)
+        card2 = Card.new(:spade, '3', 3)
+        card3 = Card.new(:heart, 'Ace', 14)
+        deck = Deck.new([card1, card2, card3])
+
+        expect(card1).to be_an_instance_of(Card)
+        expect(deck).to be_an_instance_of(Deck)
+    end
```
abdulredd, line 15: This test doesn't have much to do with the `Player` class and would probably be better in the spec files belonging to the class that they are testing.

spec/turn_spec.rb
```diff
@@ -0,0 +1,258 @@
...
+        deck1 = Deck.new([card1, card2, card5, card8])
+        deck2 = Deck.new([card3, card4, card6, card7])
+
+        player1 = Player.new("Megan", deck1)
+        player2 = Player.new("Aurora", deck2)
+
+        turn = Turn.new(player1, player2)
+        expect(card1).to be_an_instance_of(Card)
+        expect(deck1).to be_an_instance_of(Deck)
+        expect(player1).to be_an_instance_of(Player)
```
abdulredd, line 28: Lines 26, 27, 28 have already been tested elsewhere. We do not need to test them again in every test.

spec/turn_spec.rb
```diff
@@ -0,0 +1,258 @@
...
+        player1 = Player.new("Megan", deck1)
+        player2 = Player.new("Aurora", deck2)
+
+        turn = Turn.new(player1, player2)
+        turn.type 
+        winner = turn.winner
+        turn.pile_cards
+        turn.award_spoils_of_war(winner)
+        expect((player1.deck.cards - [card2, card5, card8, card1, card3]).empty?).to eq(true)
+        expect((player2.deck.cards - [card4, card6, card7]).empty?).to eq(true)
```
abdulredd, line 162: I do not understand the expectation on lines 161 and 162. If `player1` wins then they should get all the cards in the spoils, so how is their deck empty? same with `player2`.

spec/turn_spec.rb
```diff
@@ -0,0 +1,258 @@
...
+        player1 = Player.new("Megan", deck1)
+        player2 = Player.new("Aurora", deck2)
+        
+        turn = Turn.new(player1, player2)
+        turn.type
+        winner = turn.winner
+        turn.pile_cards
+        turn.award_spoils_of_war(winner)
+        expect((player1.deck.cards - [card8]).empty?).to eq(true)
+        expect((player2.deck.cards - [card7, card1, card4, card2, card3, card5, card6]).empty?).to eq(true)
```
abdulredd, line 209: What are lines 208 and 209 doing? I'm very curious.

spec/turn_spec.rb
```diff
@@ -0,0 +1,258 @@
...
+
+        turn = Turn.new(player1, player2)
+        turn.type 
+        winner = turn.winner
+        turn.pile_cards
+        turn.award_spoils_of_war(winner)
+        expect((player1.deck.cards - [card2, card5, card8, card1, card3]).empty?).to eq(true)
+        expect((player2.deck.cards - [card4, card6, card7]).empty?).to eq(true)
+        expect(turn.spoils_of_war).to eq([])
+    end
```
abdulredd, line 164: Where is the expectation that checks that the `spoils_of_war` array actually contains the cards that we expect it to contain?

abdulredd (commented):
Overall, this was a great project and I really enjoyed playing with it and going through it. I did not see a `card_spec.rb` file included with your project, though. 

All in all, great job Bryan, keep up the great work!

PR #323, opened 2022-08-28
Emily Port

eport01:


spec/turn_spec.rb
```diff
@@ -0,0 +1,129 @@
+require 'rspec'
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+require './lib/game'
+
+RSpec.describe Deck do
```
abdulredd, line 8:
```ruby
RSpec.describe Turn do
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    card8 = Card.new(:diamond, '2', 2) 
+    deck1 = Deck.new([card1, card2, card5, card8])  
+    deck2 = Deck.new([card3, card4, card6, card7]) 
+    player1 = Player.new('Megan', deck1)
+    player2 = Player.new('Aurora', deck2)
+    expect(player1.name).to eq('Megan')
+
+    turn = Turn.new(player1, player2)     
+
+  end
```
abdulredd, line 26: The `it` block wants to test `Turn` initialization but we, unfortunately, do not have an expectation in here that does that. `expect(player1.name).to eq('Megan')` has already been tested in the `Player` spec so we are ok not testing it again here.

The assertions/expectations I would have liked to see here are:

```ruby
expect(turn).to be_an_instance_of(Turn)
expect(turn.player1).to eq(player1)
expect(turn.player2).to eq(player2)
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    deck2 = Deck.new([card4, card3, card6, card7])  
+    player1 = Player.new('Megan', deck1)
+    player2 = Player.new('Aurora', deck2)
+
+    turn = Turn.new(player1, player2) 
+
+    game1= Game.new(turn)
+    
+  end
+end
```
abdulredd, line 128: The final 3 `it` blocks in your `turn_spec.rb` don't have any assertions/expectations.

abdulredd (commented):
Great job on your project Emily. I believe your biggest area of opportunity is with your tests. I know how hard testing can be when you've just been introduced to it so it's not a big deal right now, keep working at it and I'm sure it will be second nature soon. 

All in all, I am very impressed. Keep up the great work!

PR #324, opened 2022-08-28
James White

James-E-White:


lib/game.rb
```diff
@@ -0,0 +1,111 @@
...
+               :full_deck,
+               :deck1,
+               :deck2
+
+     def initialize
+      @full_deck = new_deck
+      @deck1 = @full_deck[0, 26]
+      @deck2 = @full_deck[26, 26]
+      @player1 = Player.new("Megan", @deck1)
+      @player2 = Player.new("Aurora", @deck2)
```
jamisonordway, line 24: It would be a better practice to create these Player objects outside of the game class, which would allow them to have a dynamic `name` attribute. As such, the Deck objects would probably need to be created outside of this class, too.

lib/game.rb
```diff
@@ -0,0 +1,111 @@
...
+      @full_deck = Deck.new(cards)
+      cards.shuffle!
+      # deck1 = Deck.new(@full_deck.cards[0, 26])
+      # deck2 = Deck.new(@full_deck.cards[26, 26])
+
+    end
+
+    def start
+      welcome = "Welcome to War! (or Peace) This game will be played with 52 cards.
+The players today are Megan and Aurora.
```
jamisonordway, line 93: With the way you currently have your code organized, is there any benefit to hard-coding 'Megan' and 'Aurora' instead of calling `@player1.name`/`@player2.name`?

lib/player.rb
```diff
@@ -0,0 +1,19 @@
...
+      @deck = deck
+    end
+
+     def has_lost?
+          if @deck.cards.length > 0
+            false
+          else
+            true
+
+      end
```
jamisonordway, line 17: Keep an eye on your blank spaces and indentation.
```
if @deck.cards.length > 0
  false
else
  true
end
```
Alternatively, you could accomplish this conditional logic on one line and get a boolean return value with something like
```
def has_lost?
  @deck.cards.empty?
end
```

lib/turn.rb
```diff
@@ -0,0 +1,82 @@
...
+      #how the game is played. must have a winner or war! if two war than spoils of war! rank_of_cards_at
+      #get discarded with no winner.
+    if @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0) && @player1.deck.rank_of_card_at(2) == @player2.deck.rank_of_card_at(2)
+      :mutually_assured_destruction
+    elsif @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
+      :war
+    elsif @player1.deck.rank_of_card_at(0) != @player2.deck.rank_of_card_at(0)
+        :basic
+
+    end
```
jamisonordway, line 22: The order of checks in this conditional branching makes a lot of sense to me-- if both conditions aren't met for mutually assured destruction, check for the war condition. if that isn't met, it must be a basic turn. Nice! 🎉

lib/turn.rb
```diff
@@ -0,0 +1,82 @@
...
+       @spoils_of_war << @player1.deck.remove_card
+       @spoils_of_war << @player2.deck.remove_card
+     elsif type == :war
+       3.times do
+         @spoils_of_war << @player1.deck.remove_card
+         @spoils_of_war << @player2.deck.remove_card
+       end
+     elsif type == :mutually_assured_destruction
+       
+     end
```
jamisonordway, line 57: This blank line feels a bit odd to me.

spec/player_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  player = Player.new("Clarissa", deck)
+  expect(player).to be_an_instance_of(Player)
+  expect(deck.cards).to eq([card1, card2, card3])
+  expect(player.has_lost?).to eq(false)
+  player.deck.remove_card
+  expect(player.has_lost?).to eq(false)
+  player.deck.remove_card
+  expect(player.has_lost?).to eq(false)
+  player.deck.remove_card
+  expect(player.has_lost?).to eq(true)
```
jamisonordway, line 35: great job testing that the `remove_card` method must be called 3 times before `player.has_lost?` can be true.

PR #325, opened 2022-08-28
Naomi Yocum

naomiyocum:


war_or_peace_runner.rb
```diff
@@ -0,0 +1,73 @@
...
+]
+
+war_deck.shuffle!
+
+deck1 = Deck.new(war_deck.take(26))
+deck2 = Deck.new(war_deck.last(26))
+
+player1 = Player.new('Megan', deck1)
+player2 = Player.new('Aurora', deck2)
+turn = Turn.new(player1, player2)
```
epintozzi, line 69: Your runner overall looks great. I probably would try to move everything except the last two lines into the `Game` class or some kind of setup class if I were going to refactor this.

lib/card.rb
```diff
@@ -0,0 +1,10 @@
+class Card
+  attr_reader :suit,
+                :value,
+                :rank
```
epintozzi, line 4: Very minor nit, but if you're going to put attr_readers on separate lines, they should all be aligned, so I'd scoot `:value` and `:rank` back two spaces to line everything up.

spec/card_spec.rb
```diff
@@ -10,7 +11,7 @@
 
   it "has readable attributes" do
     card = Card.new(:diamond, 'Queen', 12)
-
+    #binding.pry
```
epintozzi, line 14: It's generally best practice to remove unused/commented out code before we make a PR. We don't want to merge unused code to main, even if it's just a test.

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
...
+    high_cards
+  end
+
+  def percent_high_ranking
+    total_high = 0.0
+    cards.each do |card|
+      if card.rank >= 11
+        total_high += 1
+      end
+    end
```
epintozzi, line 28: This code is almost identical to the code in your `high_ranking_cards` method. Instead, I would set your `total_high` variable on line 23 equal to `high_ranking_cards` and count that instead.

It would look something like this:
```ruby
  def percent_high_ranking
    total_high = high_ranking_cards.count.to_f
    ((total_high / @cards.length) * 100).round(2)
  end
```

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def high_ranking_cards
+    high_cards = []
+    cards.each do |card|
+      if card.rank >= 11
+        high_cards << card
+      end
+    end
+    high_cards
```
epintozzi, line 19: This looks great! When you're ready for another challenge, see if you can solve this with `select` instead of `each`.

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
+class Deck
+  attr_reader :cards
+
+  def initialize(cards)
+    @cards = cards
+  end
+
+  def rank_of_card_at(index)
+    cards[index].rank
```
epintozzi, line 9: The rule of thumb is we want to use the instance variable in the class where it's defined (rather than the method). Both ways work, it's just convention to use the instance var.
```suggestion
    @cards[index].rank
```

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def high_ranking_cards
+    high_cards = []
+    cards.each do |card|
```
epintozzi, line 14:
```suggestion
    @cards.each do |card|
```

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
...
+      if card.rank >= 11
+        high_cards << card
+      end
+    end
+    high_cards
+  end
+
+  def percent_high_ranking
+    total_high = 0.0
+    cards.each do |card|
```
epintozzi, line 24:
```suggestion
    @cards.each do |card|
```

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
...
+      if card.rank >= 11
+        total_high += 1
+      end
+    end
+    ((total_high / cards.length) * 100).round(2)
```
epintozzi, line 29:
```suggestion
    ((total_high / @cards.length) * 100).round(2)
```

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def remove_card
+    cards.shift
```
epintozzi, line 33:
```suggestion
    @cards.shift
```

lib/deck.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def add_card(card)
+    cards << card
```
epintozzi, line 37:
```suggestion
    @cards << card
```

lib/turn.rb
```diff
@@ -0,0 +1,66 @@
...
+      (player1.deck.rank_of_card_at(2) == player2.deck.rank_of_card_at(2))
+      :mutually_assured_destruction
+    elsif player1.deck.rank_of_card_at(0) == player2.deck.rank_of_card_at(0)
+      :war
+    end
+  end
+
+  def pile_cards
+    if type == :player1_lost
+      spoils_of_war.concat(player1.deck.cards)
```
epintozzi, line 27: Ideally we should be able to use our `award_spoils` method instead of manually concatenating here (and on line 30) .

lib/turn.rb
```diff
@@ -0,0 +1,66 @@
...
+      player1.deck.cards.clear
+    elsif type == :player2_lost
+      spoils_of_war.concat(player2.deck.cards)
+      player2.deck.cards.clear
+    elsif type == :mutually_assured_destruction
+      3.times do
+        player1.deck.cards.shift
+        player2.deck.cards.shift
```
epintozzi, line 35: Can we use our `remove_card` method from `Deck` here instead?

lib/turn.rb
```diff
@@ -0,0 +1,66 @@
...
+      player2
+    elsif type == :basic && player1.deck.rank_of_card_at(0) > player2.deck.rank_of_card_at(0)
+      player1
+    elsif type == :basic && player1.deck.rank_of_card_at(0) < player2.deck.rank_of_card_at(0)
+      player2
+    end
+  end
+
+  def award_spoils(winner)
+    winner.deck.cards.concat(spoils_of_war.shuffle!)
```
epintozzi, line 63: Can we use our `add_card` method from `Deck` here instead?

lib/start.rb
```diff
@@ -0,0 +1,57 @@
+class Game
```
epintozzi, line 1: This file should be named `game.rb` to match the class name `Game`.

lib/start.rb
```diff
@@ -0,0 +1,57 @@
...
+      elsif @turn.type == :war
+        winner = @turn.winner
+        @turn.pile_cards
+        p "Turn #{counter}: WAR - #{winner.name} won 6 cards"
+        @turn.award_spoils(winner)
+      elsif @turn.type == :basic
+        winner = @turn.winner
+        @turn.pile_cards
+        p "Turn #{counter}: #{winner.name} won 2 cards"
+        @turn.award_spoils(winner)
```
epintozzi, line 42: There's a lot of repetition happening in lines 34-37 and 39-42. When we see ourselves repeating the same code more than once, that's a sign we should try to pull it out to its own method.

PR #326, opened 2022-08-28
Darby Newman

DarbySmith:


lib/deck.rb
```diff
@@ -0,0 +1,45 @@
+class Deck 
+  attr_reader :high_ranking_card, :percent_high_ranking
+  attr_accessor :cards, :rank_of_card_at
```
jamisonordway, line 3: To speak to your question about attr_reader vs attr_accessor, `attr_accessor` is only necessary if you want the ability to **read** as well as **write** the value of the instance variable it exposes, from outside of the class. Usually, an `attr_reader` is sufficient.
If you haven't already, try running your tests and `war_or_peace_runner.rb` with all of these defined as `attr_reader` and see what happens.

lib/deck.rb
```diff
@@ -0,0 +1,45 @@
...
+  end 
+
+  def high_ranking_card
+    high_cards = []
+    @cards.each do |card| 
+      if card.rank >= 11
+        high_cards << card
+      end 
+    end
+    high_cards
```
jamisonordway, line 20: This works! Is there another enumerable you could use instead of `each`, so that you wouldn't need the `high_cards` accumulator variable?

lib/deck.rb
```diff
@@ -0,0 +1,45 @@
...
+    # @cards.each do |card| 
+    #   if card.rank >= 11
+    #     high_cards << card
+    #   end 
+    # end
+
+    # percent = ((high_cards.length.to_f / @cards.length.to_f) * 100).to_i
+    # percent
+
+    percent = ((high_ranking_card.length.to_f / @cards.length.to_f) * 100).to_i
```
jamisonordway, line 34: Nice job refactoring this code to be more DRY 🎉

lib/turn.rb
```diff
@@ -0,0 +1,88 @@
...
+      :basic
+    elsif @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
+      if @player1.deck.cards.length < 3 || @player2.deck.cards.length < 3
+        :war
+      elsif @player1.deck.rank_of_card_at(2) == @player2.deck.rank_of_card_at(2)
+        :mutually_assured_destruction
+      elsif @player1.deck.rank_of_card_at(2) != @player2.deck.rank_of_card_at(2)
+        :war
+      end
+    end
```
jamisonordway, line 27: Here's a human-readable summary of the conditions we want to check for:
- :basic means player 1 and player 2 do NOT have the same rank_of_card_at(0) values
- :war means both players’ rank_of_card_at(0) are the same
- :mutually_assured_destruction means both players' rank_of_card_at(0) AND both players' rank_of_card_at(2) are the same

Imagine a world where we have lots of helper methods to break this down:
```
# this method returns 'true' when both players’ rank_of_card_at(0) AND rank_of_card_at(2) are the same.
def mutually_assured_destruction? 
  (@player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)) &&
    (@player1.deck.rank_of_card_at(2) == @player2.deck.rank_of_card_at(2))
end
```
```
# this method returns 'true' when both players’ rank_of_card_at(0) are the same
def war?
  @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0)
end
```
```
if mutually_assured_destruction? # war and basic conditions would both evaluate to false here
  :mutually_assured_destruction
elsif war?
  :war
else # if it's not mutually assured destruction, and it's not war, it must be basic
  :basic
end
```

spec/game_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'rspec'
+require 'pry'
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+require './lib/game'
+
+RSpec.describe Game do
```
jamisonordway, line 9: To address your question in the submission form: normally we would want to test every method in a class; this is the exception since we did not require you to test methods that require command line input/output

PR #327, opened 2022-08-28
Bryan Flanagan War or Peace

bflanagan138:


war_or_peace_runner.rb
```diff
@@ -0,0 +1,79 @@
+class Start
```
epintozzi, line 1: Our runner shouldn't be defining a class. If we want a class to hold these behaviors, it should go in the lib directory. Our runner shouldn't have very much code in it. Probably just a couple lines to actually start the game.

war_or_peace_runner.rb
```diff
@@ -0,0 +1,79 @@
...
+    attr_reader
+    attr_accessor
```
epintozzi, line 3: If we aren't defining any `attr_readers` or `attr_accessors` we don't need to list these here.

war_or_peace_runner.rb
```diff
@@ -0,0 +1,79 @@
...
+		card44 = Card.new(:club, '6', 6)
+		card45 = Card.new(:club, '7', 7)
+		card46 = Card.new(:club, '8', 8)
+		card47 = Card.new(:club, '9', 9)
+		card48 = Card.new(:club, '10', 10)
+		card49 = Card.new(:club, 'Jack', 11)
+		card50 = Card.new(:club, 'Queen', 12)
+		card51 = Card.new(:club, 'King', 13)
+    	card52 = Card.new(:club, 'Ace', 14)
+    end
```
epintozzi, line 70: If we put these cards in an array, we can shuffle them before divvying them out to players. If we don't shuffle every game will have the exact same outcome every time we play.

lib/turn.rb
```diff
@@ -0,0 +1,71 @@
+class Turn
+    attr_reader :player1, :player2, :spoils_of_war, :winner
```
epintozzi, line 2:
```suggestion
    attr_reader :player1, :player2, :spoils_of_war
```

We don't need or want a `winner` `attr_reader` because we are defining a method for determining a winner on line 20. Remember an `attr_reader` is a shortcut for creating a method that allows us to read the value of an instance variable. If we define methods by the same name more than once, it can cause conflicts.

lib/turn.rb
```diff
@@ -0,0 +1,71 @@
...
+    def initialize(player1, player2)
+      @player1 = player1
+      @player2 = player2
+      @spoils_of_war = []
+      @winner = winner
```
epintozzi, line 7: We won't want to instantiate with a `winner` because we calculate that after the turn is complete. The `winner` method on line 20 is sufficient to give us what we need.

lib/turn.rb
```diff
@@ -0,0 +1,71 @@
...
+          :war 
+        end
+    end
+
+  def winner
+        if type == :basic
+            if player1.deck.rank_of_card_at(0) >= player2.deck.rank_of_card_at(0)
+                @winner = @player1
+            else
+                @winner = @player2
```
epintozzi, line 25: Since I've suggested we don't need an instance var for `winner`, we can just return the winning player directly without assigning them to a variable. Same thing on lines 30 and 32.

lib/turn.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    def pile_mutually_assured_destruction
+        3.times { @player1.deck.remove_card }
+        3.times { @player2.deck.remove_card }
+    end
+
+    def award_spoils
+        # binding.pry
+        @spoils_of_war.each do |spoils|
+            @winner.deck.add_card(spoils)
```
epintozzi, line 68: Nice job! I'm glad you thought to use the `add_cards` method you wrote for `Deck`

lib/game.rb
```diff
@@ -0,0 +1,9 @@
+class Start
```
epintozzi, line 1: It seems like maybe you started combining your intentions for this class with the runner. I'd pull most of the code from the runner into this class. Whatever your class name is should match the file name. So if you call this class `Start` you should name the file `start.rb`. I personally would probably keep the file name as `game.rb` and change the class name to `Game` because that more accurately describes the purpose of the class.

lib/turn.rb
```diff
@@ -0,0 +1,71 @@
...
+        elsif player1.deck.rank_of_card_at(0) != player2.deck.rank_of_card_at(0)
+            :basic
+        elsif player1.deck.rank_of_card_at(0) == player2.deck.rank_of_card_at(0)
+          :war 
+        end
+    end
+
+  def winner
+        if type == :basic
+            if player1.deck.rank_of_card_at(0) >= player2.deck.rank_of_card_at(0)
```
epintozzi, line 22:
```suggestion
            if player1.deck.rank_of_card_at(0) > player2.deck.rank_of_card_at(0)
```
I think we want this just to be a `>`, not `>=` because if it were equal it would be a war situation. Same for line 29.

PR #328, opened 2022-08-28
AetherKerstens_War_of_peace

aether-kerstens:
I think its good? It said there were no failures and the deadline is coming close


spec/turn_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      turn.type
+      winner = turn.winner
+      turn.pile_cards
+      turn.spoils_of_war
+      turn.award_spoils(winner)
+      player1.deck
+      player2.deck
+    end
+  end
+  describe ':basic' do
```
jamisonordway, line 38: great job using describe blocks to organize the different turn type scenarios. Just keep an eye on your indentation.

spec/turn_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+  expect(@turn.spoils_of_war).to eq([])
+  expect(@player1.deck.cards).to eq([@card8])
+  expect(@player2.deck.cards).to eq([@card7])
+end
+
+it "#award_spoils :mutually_assured_destruction" do
+  @turn.mutually_assured_destruction_card_pile
+  expect(@player1.deck.cards).to eq([@card8])
+  expect(@player2.deck.cards).to eq([@card7])
+end
```
jamisonordway, line 142: re: indentation -- these 'it' blocks are inside of the `describe ':mutually_assured_destruction' do` block, so they should be indented at the same level as the `before(:each)` on line 111

spec/deck_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+  describe  "#percent_high_ranking" do
+    it "return percentage of cards that are high" do
+      card1 = Card.new(:diamond, 'Queen', 12)
+      card2 = Card.new(:spade, '3', 3)
+      card3 = Card.new(:heart, 'Ace', 14)
+      cards = [card1, card2, card3]
+      deck = Deck.new(cards)
+      high_rank = [card1, card3]
+      binding.pry
```
jamisonordway, line 48: don't forget to remove your `binding.pry` when you aren't using them.

PR #329, opened 2022-08-28
War or Peace project

msakr21:
Last commit has added files for trying to work on iteration 4 which could be a source of noise. Penultimate commit has fully functioning code for iteration 3.


lib/deck.rb
```diff
@@ -0,0 +1,56 @@
...
+        high_rank_cards = []
+
+        @cards.each do |card|
+            if card.rank >= 11
+                high_rank_cards << card
+            end
+        end
+
+        high_rank_cards
+
```
epintozzi, line 30: This works great! When you're ready for a refactor, see if you can solve this using the `select` enumerable method instead.

lib/deck.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+
+    def percent_high_ranking
+
+        
+        numerator = high_ranking_cards.length() * 1.00
+
+        denominator = @cards.length() * 1.00
```
epintozzi, line 39:
```suggestion
        numerator = high_ranking_cards.length.to_f

        denominator = @cards.length.to_f
```

In Ruby we don't need to include empty parens when there are no arguments. It is convention to these those off. It's also preferred to use `.to_f` which converts an integer to a float instead of multiplying by 1.00

lib/dump.txt
```diff
@@ -0,0 +1,58 @@
+#100000000.times do
```
epintozzi, line 1: It's good practice not to include commented out/unused code in a PR to main. We only want production ready code merged in.

lib/turn.rb
```diff
@@ -0,0 +1,163 @@
...
+            TRUE
+        elsif @player1.deck.cards.length < 3 
+            TRUE
+        elsif @player2.deck.cards.length < 3 
+            TRUE
+        elsif @player2.deck.cards.length < 3 
+            TRUE
+        else
+            FALSE
+        end
```
epintozzi, line 89: Lines 79 and 81 as well as 83 and 85 are duplicates. We only need to check a condition once. This could be shortened to:
```ruby
  def not_enough_cards?
    if @player1.deck.cards.length < 3
      TRUE
    elsif @player2.deck.cards.length < 3
      TRUE
    else
      FALSE
    end
```

or this works as well:
```ruby
    if @player1.deck.cards.length < 3 || @player2.deck.cards.length < 3
      TRUE
    else
      FALSE
    end
```

lib/turn.rb
```diff
@@ -0,0 +1,163 @@
...
+
+    end
+
+    def not_enough_cards
```
epintozzi, line 93: This method name is a little confusing to me, especially with `not_enough_cards?` just above it. I think I'd rename it something like `winner_when_not_enough_cards` since what it is actually doing is returning us a winner.

lib/turn.rb
```diff
@@ -0,0 +1,163 @@
...
+
+            
+
+                if winner.deck.cards.empty? == FALSE
+
+                    i = @spoils_of_war.length
+
+                    i.times do
+                    winner.deck.cards << @spoils_of_war.shift
+                    end
```
epintozzi, line 150: Ruby gives us `each` which would be a preferred implementation here. Generally, ruby convention prefers using our built in enumerable methods over looping with a counter.

spec/turn_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+
+
+        player1 = Player.new("Megan", deck1)
+        player2 = Player.new("Aurora", deck2)
+
+        turn = Turn.new(player1, player2)
+
+        expect(turn.type).to eq(:mutually_assured_destruction)
+
+        expect(turn.winner).to eq(nil)
```
epintozzi, line 230: The project requirements indicate this should actually return "No Winner". I could see a lot of reasons why `nil` makes more sense, but technically this is not built to spec.

spec/turn_spec.rb
```diff
@@ -0,0 +1,375 @@
...
+
+        expect(turn.type).to eq(:basic)
+
+        turn.pile_cards
+
+        expect(player1.deck.cards).to eq([card2, card5, card8])
+        expect(player2.deck.cards).to eq([card4, card6, card7])
+        
+        expect(turn.spoils_of_war).to include(card1, card3)
+        #^include works better for the following tests to avoid the need to worry about card order while checking
```
epintozzi, line 264: I agree. If we don't NEED to care about the order, I also prefer includes for this reason.

lib/game.rb
```diff
@@ -0,0 +1,240 @@
+require './lib/card'
+require './lib/deck'
+require './lib/player'
+require './lib/turn'
+
+
+
+class Game
```
epintozzi, line 8: All the extra line breaks make this class really difficult to read through.

lib/game.rb
```diff
@@ -0,0 +1,240 @@
...
+
+
+        create_high_value(diamonds, :diamond)
+        create_high_value(hearts, :heart)
+        create_high_value(spades, :spade)
+        create_high_value(clubs, :club)
+
+
+        full_deck_cards = diamonds + hearts + spades + clubs
+        full_deck_cards = full_deck_cards.shuffle
```
epintozzi, line 78: I like the shuffle. Without a shuffle we end up playing the same game every time.

lib/game.rb
```diff
@@ -0,0 +1,240 @@
...
+        end
+
+    end
+
+
+
+
+
+
+    def start
```
epintozzi, line 142: This method feels very long to me. Similar to how you pull your `welcome_message` method out to its own method in the `welcome_loop`, I wish you'd done more of that here so your `start` method would be a little more digestible. There's a lot going on and that leaves us prone to bugs.

lib/card_generator.rb
```diff
@@ -0,0 +1,62 @@
...
+require './lib/game'
+
+require 'pry'
+
+class CardGenerator
+
+    attr_reader :deck1,
+                :deck2
+
+    def initialize(file_name)
```
epintozzi, line 14: This looks great! I do think this class is doing too much though. I'd have stopped around line 37 after creating all the cards. Then, your `Game` can still be responsible for shuffling and divvying out the cards. The `CardGenerator`'s only job should be generating cards. I'd also like to have seen tests for this class.

epintozzi (commented):
Ruby convention is that every line that needs to be indented should be indented two spaces. It looks like yours is 4. Double check your editor is set that if you do a tab it's only 2 spaces and if you are manually indenting, pull it back just a bit 😉 

We also don't need line breaks in between every line. Convention would only play a line break in between methods, but not between the lines within our methods.

lib/deck.rb
msakr21: This comment is somewhat confusing for me: do you mean for me to revisit this to try to refactor later on? Or is this something you would like me to do for the assessment?

I'm going to make an educated guess that it's the prior, and that it's for practice, correct?

lib/deck.rb
msakr21: I'd realized that it wasn't needed later on in my codes but missed removing it here. The to_f method is something I got more exposure to after the project, so this makes a lot of sense now and definitely looks cleaner to me!

lib/turn.rb
msakr21: I think I accidentally copy pasted that if condition and didn't realize it! Would you say it's better practice to keep conditions on a single line if they can be merged because of the restricting methods to 7 lines requirement that we'll start having later on?

lib/turn.rb
msakr21: Is there a technical reason for that, or is more of a community agreed upon thing?

spec/turn_spec.rb
msakr21: Yeah. At the time, I wasn't entirely clear on whether what needed to be returned was a print of "No Winner" or "No Winner" itself, and I didn't know at the time that I could plug in print "No Winner" as a test, so I had a different part of my code handle printing out this scenario. I also recall running into problems when I tried to use the return value of this later while running the game.

I thought the pry commands and outputs were more of an example rather than needing to be followed literally.

lib/game.rb
msakr21: I'm definitely going to need to work on my getting claustrophobic when things are bunched together more and avoid spacing things out!

lib/game.rb
msakr21: I actually realized that post submission when I heard about keeping methods to 7 lines at most (can't remember who I heard it from, I think it was Abdul?) later on, and how having more helper methods can allow us to test for each individual block of code and avoid running into the user input testing issue that I experienced.

msakr21:
Thanks for all the feedback, Erin, I learned a lot from this!

lib/game.rb
epintozzi: The rule of thumb is about 7 yeah. Essentially once our methods start getting longer than that, it's a sign that the method might be doing too much. It's not an absolute, just a trigger to take a second look. Sometimes it makes sense to have a couple longer methods for the sake of readability, for example.  But yes, helpers allows us to test things a little more granularly which can prevent bugs as well.

lib/turn.rb
epintozzi: I would say it's better practice to combine them, not because of the 7 line limit but because the fewer paths and outcomes we have, the fewer chances for bugs. You could also move that condition to a helper method which would make the whole thing a little easier to read. Something like:
```ruby
def player_has_fewer_than_3_cards?
  @player1.deck.cards.length < 3 || @player2.deck.cards.length < 3
end
```

And then in the `if` statement you could write:
```ruby
if player_has_fewer_than_3_cards?
  true
else
  false
end
```

lib/turn.rb
epintozzi: Community agreed upon thing. Ruby has a ton of built in enumerable methods and it's pretty much always preferable to use a built in method over a counter/loop if we can.

lib/deck.rb
epintozzi: Yes just for practice if you feel like coming back to it!

PR #330, opened 2022-08-28
Annie Pulzone - War_or_Peace Project Pull Request

ajpulzone:


lib/deck.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def percent_high_ranking
+    percent = ((high_ranking_cards.length.to_f / @cards.length.to_f) * 100).to_i
+    percent
+  end
+
+  def remove_card
+    cards.shift
+  end
```
abdulredd, line 29: I noted in all your other methods you use the instance variable `@cards` but in this method, you use `cards`, any reason for the change here?

lib/turn.rb
```diff
@@ -0,0 +1,78 @@
...
+  end
+
+  def pile_cards
+      if type == :basic
+      #each player will send one card (the top card) to the spoils pile
+      # spoils_of_war = []
+      spoils_of_war << player1.deck.remove_card
+      spoils_of_war << player2.deck.remove_card
+      # spoils_of_war
+      end
```
abdulredd, line 47: Lines 44 & 45 need to be indented one space over but other than that great job using the `remove_card` method that we have defined on the `Deck` class. Another great example of keeping your code DRY!

It's also recommended to remove all your comments when submitting your work. If you want to keep a copy branch that has all the comments on it for your own benefit, that is definitely an option that you can take.

lib/turn.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      if type == :mutually_assured_destruction
+        player1.deck.remove_card
+        player1.deck.remove_card
+        player1.deck.remove_card
+        player2.deck.remove_card
+        player2.deck.remove_card
+        player2.deck.remove_card
+      end
```
abdulredd, line 65: The final 2 `if` blocks have quite a bit of repetition, what could we have done here instead to make this more DRY? Take a look at the `#times` method for clues.

lib/turn.rb
```diff
@@ -0,0 +1,78 @@
...
+    #
+    # elsif type == :mutually_assured_destruction
+    #   #each player will remove three cards from play (the top three cards in their deck).
+    #   #These cards are NOT send to the spoils pile, they are simply removed from each players deck
+  end
+
+  def award_spoils
+    @winner.deck.cards.concat(spoils_of_war)
+    #research concat
+    end
```
abdulredd, line 77: This method is wonderfully written and works as intended if the game's are of a `:basic` or `:war` type. If the game is `:mutually_assured_destruction` the game breaks here. This is because, if you remember, when a game is `:mutually_assured_destruction` we get `"No Winner"` returned from the `winner` method. And therefore when we call `award_spoils` and it goes to get `@winner`, it will have the value `"No Winner"` string and you cannot call `.deck` on a String.

One way to avoid this behavior is to add a conditional statement to this method so that way it only runs if an actual player object is stored in `@winner`.

```ruby
def award_spoils
  if @winner != "No Winner" #<== If winner is not equal to "No Winner"
    @winner.deck.cards.concat(spoils_of_war) #<== then do this
  end #<== otherwise do nothing
end
```

===

One other thing. According to the interaction pattern this method is supposed to take in an argument but your method doesn't. I imagine the method would still work as written using the `@winner` instance variable, but we want to make sure that when we are given a specification like the interaction, we try and stick to it exactly. 🙂 

```ruby
def award_spoils(winner)
  # code...
end
```

lib/game.rb
```diff
@@ -0,0 +1,124 @@
...
+
+def initialize(turn)
+  @full_deck = full_deck
+  @deck1 = @full_deck[0, 26]
+  @deck2 = @full_deck[26, 26]
+  @player1 = Player.new("Perrin", @deck1) #Player.new("Perrin", deck1)
+  @player2 = Player.new("Egwene", @deck2) #Player.new("Egwene", deck2)
+  @cards = cards
+  @current_turn = current_turn
+end
```
abdulredd, line 17: In your `#initialize` method, I couldn't find where `full_deck` on line 10, `cards` on line 15, and `current_turn` on line 16 were coming from.

Here is how I refactored this method:

```ruby
def initialize
  @full_deck = create_deck #<== calling the 'create_deck' method to get my cards
  @deck1 = @full_deck[0, 26]
  @deck2 = @full_deck[25, 26]
  @player1 = Player.new("Perrin", @deck1)
  @player2 = Player.new("Perrin", @deck2)
end
```

You'll notice that I removed `@cards = cards` and `@current_turn = current_turn` because I didn't see a `cards` or `current_turn` method inside this class.

lib/game.rb
```diff
@@ -0,0 +1,124 @@
...
+  cards << card45 = Card.new(:club, "7", 7)
+  cards << card46 = Card.new(:club, "8", 8)
+  cards << card47 = Card.new(:club, "9", 9)
+  cards << card48 = Card.new(:club, "10", 10)
+  cards << card49 = Card.new(:club, "Jack", 11)
+  cards << card50 = Card.new(:club, "Queen", 12)
+  cards << card51 = Card.new(:club, "King ", 13)
+  cards << card52 = Card.new(:club, "Ace", 14)
+
+  full_deck = Deck.new(cards)
```
abdulredd, line 86: This line, `full_deck = Deck.new(cards)`, can be safely removed and once all the cards are created and added to the `cards` array and shuffled, that array can be returned to `@full_deck` in the `initialize`, where it can then be split between both players.

spec/deck_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  describe '#percent_high_ranking' do
+    it "will return the percent (interger) of cards in deck with a rank >= 11" do
+      card1 = Card.new(:diamond, "Queen", 12)
+      card2 = Card.new(:spade, "3", 3)
+      card3 = Card.new(:heart, "Ace", 14)
+      cards = [card1, card2, card3]
+      deck = Deck.new(cards)
+      deck.percent_high_ranking
+      expect(deck.percent_high_ranking).to eq(66)
```
abdulredd, line 65:
```ruby
expect(deck.percent_high_ranking).to eq(66.67)
```

spec/deck_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      cards = [card1, card2, card3]
+      deck = Deck.new(cards)
+      expect(deck.cards).to eq([card1, card2, card3])
+      card4 = Card.new(:club, "5", 5)
+      deck.add_card(card4)
+      expect(deck.cards).to eq([card1, card2, card3, card4])
+  end
+  end
+
+end
```
abdulredd, line 98: Wow, I really like all the tests here. They test exactly what they should and they are organized and set up in a way that is easy to read and understand!

spec/player_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      player.deck.remove_card
+      expect(player.has_lost?).to eq(false)
+      player.deck.remove_card
+      expect(player.has_lost?).to eq(false)
+      player.deck.remove_card
+      expect(player.has_lost?).to eq(true)
+    end
+  end
+
+end
```
abdulredd, line 39: Another great test suite!

spec/turn_spec.rb
```diff
@@ -0,0 +1,294 @@
...
+      turn.pile_cards
+      expect(turn.spoils_of_war).to eq([])
+      expect(player1.deck.cards.length).to eq(1)
+      expect(player1.deck.cards).to eq([card8])
+      expect(player2.deck.cards.length).to eq(1)
+      expect(player2.deck.cards).to eq([card7])
+
+    end
+  end
+end
```
abdulredd, line 294: Your `Turn` tests are awesome! The only one you missed was testing `award_spoils` for a `:mutually_assured_destruction` scenario. If you had added that test, you would have realized it was a point in your program that had the potential for breaking and would have probably added that condition I suggested earlier.

war_or_peace_runner.rb
```diff
@@ -0,0 +1,35 @@
+require "./lib/game.rb"
+
+game = Game.new
```
abdulredd, line 3: In your `Game` classes initialize method, you are expecting an argument of `(turn)` but you never pass that in here when initializing a new `Game` object. I'm not sure if passing in a `Turn` object was the intention or not but just something I wanted to point out. 

When I tried running `ruby war_or_peace_runner.rb` I was getting `game.rb:9:in 'initialize': wrong number of arguments (given 0, expected 1) (ArgumentError)`.

abdulredd (commented):
I like the direction that your `Game` class was taking and I am sure if you had another 24 hours you would have been able to implement more of the ideas that you had.

It's obvious you know exactly what you are doing and if I were to pick your biggest strength it would definitely be your tests, they were amazing!

Keep up the great work Annie!

PR #331, opened 2022-08-28
Kenz Leng - war or peace submission

kenzjoy:


lib/turn.rb
```diff
@@ -0,0 +1,104 @@
...
+            #  spoils_of_war << player1.deck.cards.shift(3)
+            #  spoils_of_war << player2.deck.cards.shift(3)
+            #  why does this code equal a nested array of 12 cards
+        else type == :mutually_assured_destruction
+            3.times do
+                player1.deck.remove_card
+                player2.deck.remove_card
+            end
+        end
+    end
```
abdulredd, line 57: I like the whole setup of this entire class. Using the `.times` method to perform the same operation multiple times, using the already defined `.remove_card` method that we previously defined on the `Deck` class, there's a lot to like here!!

lib/turn.rb
```diff
@@ -0,0 +1,104 @@
...
+            spoils_of_war << player1.deck.remove_card
+            spoils_of_war << player2.deck.remove_card
+        elsif type == :war
+            3.times do
+                spoils_of_war << player1.deck.remove_card
+                spoils_of_war << player2.deck.remove_card
+            end
+            #  spoils_of_war << player1.deck.cards.shift(3)
+            #  spoils_of_war << player2.deck.cards.shift(3)
+            #  why does this code equal a nested array of 12 cards
```
abdulredd, line 50: A housekeeping tip: When turning in work, it's always good practice to remove all comments out of your work.

lib/turn.rb
```diff
@@ -0,0 +1,104 @@
...
+        end
+        
+        if player1.deck.cards.length == 0
+            puts "*~*~*~* #{player2.name} has won the game! *~*~*~*"
+        elsif player2.deck.cards.length == 0
+            puts "*~*~*~* #{player1.name} has won the game! *~*~*~*"
+        else
+            puts "---- DRAW ----"
+        end
+    end
```
abdulredd, line 103: Although this `start` method works just fine where it is defined in the `Turn` class, I would have liked to see it inside another class like iteration 3 asked for.  I saw you beginning to flesh out that class so I'm sure that's where you were headed anyway.

war_or_peace_runner.rb
```diff
@@ -0,0 +1,83 @@
...
+    
+    deck1 = full_deck[0..25]
+    deck2 = full_deck[26..51]
+
+    player1 = Player.new('Megan', deck1)
+    player2 = Player.new('Aurora', deck2)
+
+    turn = Turn.new(player1, player2)
+
+    turn.start
```
abdulredd, line 83: All the code in this file besides the `require`'s are indented over one space I almost thought they were nested inside a block or method or something. To reduce the confusion, let's try to stick to proper indentation practice and in this case that would mean un-indenting all these lines.

war_or_peace_runner.rb
```diff
@@ -0,0 +1,83 @@
...
+            card16, card17, card18, card19, card20, card21, card22, 
+            card23, card24, card25, card26, card27, card28, card29, 
+            card30, card31, card32, card33, card34, card35, card36,
+            card37, card38, card39, card40, card41, card42, card43, 
+            card44, card45, card46, card47, card48, card49, card50, 
+            card51, card52]
+
+    3.times do
+        full_deck.shuffle!
+    end
```
abdulredd, line 73: Wooh, shuffling that deck 3 times. Making sure that thing is nice a mixed! 🔀

spec/deck_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    describe 'remove_card' do 
+        it 'removes the top card from the deck' do
+            card1 = Card.new(:diamond, 'Queen', 12)
+            card2 = Card.new(:spade, '3', 3)
+            card3 = Card.new(:heart, 'Ace', 14)
+            cards = [card1, card2, card3]
+
+            deck = Deck.new(cards)
+            expect(deck.remove_card).to eq(card1)
+        end
```
abdulredd, line 65: I would have like to see another expectation here after the card has been removed to check the decks cards again to verify that the card was removed.

```ruby
expect(deck.cards).to eq([card2, card3])
```

spec/deck_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+            card2 = Card.new(:spade, '3', 3)
+            card3 = Card.new(:heart, 'Ace', 14)
+            cards = [card1, card2, card3]
+        
+            deck = Deck.new(cards)
+            card4 = Card.new(:club, '5', 5)
+            deck.add_card(card4)
+            expect(deck.cards).to eq([card1, card2, card3, card4])
+        end
+    end
```
abdulredd, line 80: If I remember correctly, in the iteration pattern, after a 4th card was added to the deck, `high_ranking_cards` and `percent_high_ranking` were checked again. These expectations should also be added to this test block.

spec/player_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+            card3 = Card.new(:heart, 'Ace', 14)
+            cards = [card1, card2, card3]
+        
+            deck = Deck.new(cards)
+
+            player = Player.new('Clarisa', deck)
+            expect(player.has_lost?).to eq(false)
+        end
+    end
+end
```
abdulredd, line 33: There are a few tests missing from this spec. We successfully check that a new `Player` instance object can be created and that we can call `has_lost?` on the `Player` object. But We don't check the `Player`'s instance variables like `.name` or `.deck`.

```ruby
expect(player.name).to eq("Clarisa")
expect(player.deck).to eq(deck)
expect(player.deck.cards).to eq(cards)
```

spec/turn_spec.rb
```diff
@@ -0,0 +1,226 @@
...
+                turn.award_spoils(player2)
+                
+                expect(player2.deck.cards).to eq([card7, card1, card4, card2, card3, card5, card6])
+            end
+        end
+    end
+
+
+    
+
```
abdulredd, line 226: I think the indentation in this file is a little "off".

The turn spec was a really thorough one and you hit a lot of the tests with flying colors. There were a few expectations missing from this spec though.

At the start of the spec, we want to make sure to add specs for reading the instance variables of a class. So that would mean:

```ruby
expect(turn.player1).to eq(player1)
expect(turn.player2).to eq(player2)
expect(turn.spoils_of_war).to eq([])
```

After that your tests are great. You have tests for all methods under a `:basic`, `:war`, and `:mutually_assured_destruction` scenario. The only one you seemed to miss was `award_spoils` for a `:mutually_assured_destruction` situation and I believe that by adding this test you would have seen the issue that I walked about before with `"No Winner"`. That test would just make sure that the `@spoils_of_war` instance variable didn't get any cards added to it and that both players had their first 3 cards removed.

abdulredd (commented):
You should be extremely proud of what you were able to put together here, I am thoroughly impressed. Your game did not initially run successfully and I had to make a few tweaks (very little actually), but then the game ran 80% of the time without issues.

The only things that were updated to make the game run were:
* Making the 2 decks in the runner file actual `Deck` objects
* Adding the `if` condition inside `award_spolis`
* Accepting any case from the user

Some lingering issues keeping the game from running flawlessly 100% of the time were taking into account a scenario where one of the players wouldn't have enough cards to enter into a `:mutually_assured_destruction` scenario and the game breaks with the error: `undefined method 'rank' for nil:NilClass`. To solve this, another condition in your `type` method needed to be added to account for this.

Great job Kenz, keep it up!

PR #336, opened 2022-08-28
Michael Marchand - PR

MarchandMD:
Hello! This was fun!


war_or_peace_runner.rb
```diff
@@ -0,0 +1,50 @@
...
+require './lib/game'
+
+# all_cards = game.create_full_deck
+# all_cards.shuffle!
+
+# player1 = Player.new('Megan', deck1)
+# player2 = Player.new('Aurora', deck2)
+
+turn = Turn.new(player1, player2)
+game = Game.new(turn)
```
epintozzi, line 17: Your `Game` class doesn't initialize with an argument so we don't need to pass `turn` here. This means we can also get rid of line 16. Your `Game` creates the players so we don't need to do that inside the runner.

war_or_peace_runner.rb
```diff
@@ -0,0 +1,50 @@
...
+
+game.start(player1, player2)
```
epintozzi, line 19: Your `start` method doesn't take arguments so we don't need to pass players here. Your `initialized` method in `Game` does that for you.

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+    @deck2 = Deck.new(full_deck)
+  end
+
+  def create_players
+    @player1 = Player.new('Megan', @deck1)
+    @player2 = Player.new('Aurora', @deck2)
+  end
+
+  def game_over?
+    @turn.player1.has_lost? || @turn.player2.has_lost?
```
epintozzi, line 77:
```suggestion
    @player1.has_lost? || @player2.has_lost?
```
I'm pretty sure you can check the player instance variable directly without relying on the turn. The players are going to be the same no matter what. Every game will have lots of turns so lets use the thing that is always there.

lib/deck.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+  end
+
+  def percent_high_ranking
+    last_digit = (high_ranking_cards.length / @cards.length.to_f * 100).to_s.slice(-1).to_i
+    if last_digit >= 5
+      (high_ranking_cards.length / @cards.length.to_f * 100).ceil(2)
+    else
+      (high_ranking_cards.length / @cards.length.to_f * 100).floor(2)
+    end
```
epintozzi, line 24: I think this is a little overcomplicated. The following should also solve the problem and is a lot more readable:
`(high_ranking_cards.count / cards.count.to_f * 100).round(2)`

lib/player.rb
```diff
@@ -0,0 +1,18 @@
...
+    @deck = deck
+    @has_lost = false
+  end
+  
+  def has_lost?
+    if @deck.cards.count.zero?
+      @has_lost = true
+    else
+      @has_lost
+    end
```
epintozzi, line 16: This feels a little complicated. I'd get rid of the `@has_lost` (and associated `attr_reader`) and just return `true` or `false` directly here.

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+
+  def split_deck
+    @deck1 = Deck.new(full_deck.slice!(0, 26))
+    @deck2 = Deck.new(full_deck)
+  end
+
+  def create_players
+    @player1 = Player.new('Megan', @deck1)
+    @player2 = Player.new('Aurora', @deck2)
+  end
```
epintozzi, line 74: You never call `split_deck` anywhere so `@deck1` and `@deck2` are `nil`. I'd probably just do it right here:
```ruby
  def create_players
    split_deck
    @player1 = Player.new('Megan', @deck1)
    @player2 = Player.new('Aurora', @deck2)
  end
```

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+  # def endgame?
+  #   turn.player1.deck.cards.length < 3 || turn.player2.deck.cards.length < 3
+  # end
+
+  # def three_card_endgame?
+  #   turn.player1.deck.cards.length == 3 || turn.player2.deck.cards.length == 3
+  # end
+
+  def two_card_endgame?
+    turn.player1.deck.cards.length == 2 || turn.player2.deck.cards.length == 2
```
epintozzi, line 89:
```suggestion
    @player1.deck.cards.length == 2 || @player2.deck.cards.length == 2
```

lib/game.rb
```diff
@@ -0,0 +1,144 @@
...
+      turn.pile_cards
+      p "turn #{counter}: #{hand_winner.name} won #{turn.spoils_of_war.count} cards"
+      turn.award_spoils(hand_winner)
+      p "Megan: #{turn.player1.deck.cards.count}| Aurora: #{turn.player2.deck.cards.count}"
+    else
+      'uh oh'
+    end
+  end
+
+  def make_turn
```
epintozzi, line 141: This method is never called anywhere so there is no turn. You probably saw a lot of 'undefined method of nil class' errors where `turn` was involved. We need to create our turn in order to call methods on it.

lib/turn.rb
```diff
@@ -0,0 +1,91 @@
...
+    when :war
+      @spoils_of_war.concat(@player1.deck.cards.shift(3), @player2.deck.cards.shift(3))
+    when :mutually_assured_destruction
+      @player1.deck.cards.slice!(0, 3)
+      @player2.deck.cards.slice!(0, 3)
+    end
+  end
+
+  def award_spoils(hand_winner)
+    hand_winner.deck.cards.concat(@spoils_of_war)
```
epintozzi, line 88: We should be able to use our `add_card` method from `Deck` here.

spec/turn_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+        # require 'pry'; binding.pry
+      end
+    end
+
+    context '#pile_cards' do
+      it 'returns the winner of the hand' do
+        turn = Turn.new(player1, player2)
+        hand_winner = turn.winner
+        expect(turn.type).to eq(:basic)
+        expect(turn.winner).to eq(player1)
```
epintozzi, line 49: We should also `expect` that the `@spoils_of_war` includes the newly piled cards.

spec/game_spec.rb
```diff
@@ -0,0 +1,339 @@
+require '../lib/game'
+require 'spec_helper'
+require 'pry'
+
+RSpec.describe Game do
```
epintozzi, line 5: Even though a lot of the test are skipped here, I really like the thought process behind the things you are planning to test. You have a good instinct for testing.

PR #337, opened 2022-08-28
Max MacGillivray

MoxieMax:


lib/turn.rb
```diff
@@ -0,0 +1,94 @@
...
+    
+    @player1 = player1
+    @player2 = player2
+    @spoils_of_war = []
+    @winner = nil
+  end
+
+  def type 
+    if player1.deck.rank_of_card_at(0) != player2.deck.rank_of_card_at(0)
+    :basic
```
jamisonordway, line 13: this applies to several places in this file; make sure the code within your `if` branches is indented properly.
```
if player1.deck.rank_of_card_at(0) != player2.deck.rank_of_card_at(0)
  :basic
elsif [other condition]
  :war
else
  :mutually_assured_destruction
end
```

lib/turn.rb
```diff
@@ -0,0 +1,94 @@
...
+      @spoils_of_war << player1.deck.remove_card
+      @spoils_of_war << player2.deck.remove_card
+    end
+    if type == :war
+      3.times {@spoils_of_war << player1.deck.remove_card}
+      3.times {@spoils_of_war << player2.deck.remove_card}
+    end
+    if type == :mutually_assured_destruction
+      3.times {player1.deck.remove_card}
+      3.times {player2.deck.remove_card}
```
jamisonordway, line 52: nifty one-liner `times` loops! 🎉

spec/player_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    deck = Deck.new([card1, card2, card3])
+    player = Player.new('Clarisa', deck)
+
+    expect(player.has_lost?).to eq(false)
+    player.deck.remove_card
+    expect(player.has_lost?).to eq(false)
+    player.deck.remove_card
+    expect(player.has_lost?).to eq(false)
+    player.deck.remove_card
+    expect(player.has_lost?).to eq(true)
```
jamisonordway, line 34: I like how this tests that `has_lost?` returns true **only** when the deck has enough cards removed.

PR #338, opened 2022-08-28
Gabe Nunez

MisterJackpots:


war_or_peace_runner.rb
```diff
@@ -0,0 +1,2 @@
+require './lib/game'
+start_game = Game.new.start
```
epintozzi, line 2: This runner looks great. Although your `Game` class doesn't appear to be done just yet, I like the decision you made to keep all the game setup pieces (like creating the decks) in that class so that all your runner needs to be responsible for is the act of starting the game.

lib/deck.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  # Method high_ranking_cards should read rank of each card in the cards array, if the
+  # card rank is greater than 10, the card should go into the high_rank_cards array
+
+
+  def high_ranking_cards
+    @high_rank_cards = []
+    @cards.each do |card|
+      if card.rank > 10
```
epintozzi, line 19: The spec indicates this should be 11 or above meaning we'd want this line to be `if card.rank >= 11`

lib/deck.rb
```diff
@@ -0,0 +1,37 @@
...
+        then @high_rank_cards << card
```
epintozzi, line 20:
```suggestion
       @high_rank_cards << card
```

lib/deck.rb
```diff
@@ -0,0 +1,37 @@
...
+    (@high_rank_cards.length.to_f / @cards.length.to_f * 100).round(2)
+  end
+
+  def remove_card
+    @cards.shift
+  end
+
+  def add_card(cards)
+    @cards.push(cards)
+  end
```
epintozzi, line 36:
```suggestion
  def add_card(card)
    @cards.push(card)
  end
```

This method only adds one card at a time, so for clarity I'd probably call this parameter card (singular) instead of cards (plural)

lib/turn.rb
```diff
@@ -0,0 +1,59 @@
...
+    else
+      3.times do
+        player1.deck.remove_card
+        player2.deck.remove_card
+      end
+    end
+  end
+
+  def award_spoils(winner)
+    winner.deck.cards.concat(@spoils_of_war)
```
epintozzi, line 57: We should be able to use our `add_card` method from `Deck` here

lib/turn.rb
```diff
@@ -0,0 +1,59 @@
...
+    elsif type == :war
+      if player1.deck.rank_of_card_at(2) > player2.deck.rank_of_card_at(2)
+        then player1
+      else player2
+      end
+
+    else type == :mutually_assured_destruction
+      "No Winner"
+    end
+  end
```
epintozzi, line 37: Overall this method is good! It works! A couple ruby style convention things I noticed to keep an eye out for:
* we don't need `then` in our `if` statements
* We should use the instance variable in the class where its defined rather than the method (so `@player1` instead of `player1`)
* We should put the action for the `else` outcome on a new line
* No need for an extra line break between our `if`, `elsif`, and `else`
Revised to follow those conventions would look something like this:

```ruby
  def winner
    if type == :basic
      if @player1.deck.rank_of_card_at(0) > @player2.deck.rank_of_card_at(0)
        @player1
      else
        @player2
      end
    elsif type == :war
      if @player1.deck.rank_of_card_at(2) > @player2.deck.rank_of_card_at(2)
        @player1
      else
        @player2
      end
    else type == :mutually_assured_destruction
      "No Winner"
    end
  end
```

lib/game.rb
```diff
@@ -0,0 +1,86 @@
+require 'pry'
+
+class Game
+  attr_reader :turn, :turn_count, :game_deck
+
+  def initialize(turn)
```
epintozzi, line 6: We probably don't need to initialize with a turn because a game will have many turns. We can create a new turn every loop rather than starting with just one.

lib/game.rb
```diff
@@ -0,0 +1,86 @@
...
+       card10, card11, card12, card13, card14, card15, card16, card17, card18,
+       card19, card20, card21, card22, card23, card24, card25, card26, card27,
+       card28, card29, card30, card31, card32, card33, card34, card35, card36,
+       card37, card38, card39, card40, card41, card42, card43, card44, card45,
+       card46, card47, card48, card49, card50, card51, card52]
+
+    5.times do
+      @game_deck.shuffle!
+    end
+  end
```
epintozzi, line 76: Great touch! I like that you shuffle before dealing the deck to ensure we get a new and random game every time we play.

PR #339, opened 2022-08-29
Sergio Azcona

Sergio-Azcona:


lib/turn.rb
```diff
@@ -0,0 +1,67 @@
...
+  def award_spoils(winner)
+    winner.deck.cards.concat(@spoils_of_war)
+  end
+
+  def strength_of_card(player, element)
+    player.deck.rank_of_card_at(element)
+  end
+
+  def cards_facing_off_index(position)
+    strength_of_card(@player1, position) > strength_of_card(@player2, position)
```
jamisonordway, line 59: I like that this method utilizes both the instance variables you have available in this scope, as well as an argument so that the 'position' logic calculated by the `strength_of_card` method is dynamic.

lib/turn.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def shovel_lost_cards_to_spoils(number_of_cards)
+    number_of_cards.times { @spoils_of_war << ( @player1.deck.remove_card) }
+    number_of_cards.times { @spoils_of_war << ( @player2.deck.remove_card) }
```
jamisonordway, line 64: I appreciate the parens on these lines, clearly communicating that it is the *return value* of `@player.deck.remove_card` that is getting shoveled into `@spoils_of_war`

PR #340, opened 2022-08-29
Will Wang

willjw2:


lib/deck.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def percent_high_ranking
+    # high_cards = 0.0
+    # @cards.each do |card|
+    #   if card.rank >= 11
+    #     high_cards += 1
+    #   end
+    # end
+    # ((high_cards / @cards.length()) * 100).round(2)
+    ((self.high_ranking_cards.length().to_f / @cards.length()) * 100).round(2)
```
jamisonordway, line 26: Great job refactoring this method to be more dry by using `high_ranking_cards` 🎉

lib/deck.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def rank_of_card_at(position)
+    @cards[position].rank
+  end
+
+  def high_ranking_cards
+    @cards.select do |card|
+      card.rank >= 11
+    end
```
jamisonordway, line 15: Awesome use of the `select` enumberable

lib/game.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  def start
+    self.display_start_message
+    user_input = gets.chomp
+    if user_input.downcase == "go"
+      self.run_game
+    else
+      self.abort("Exiting program...")
+    end
```
jamisonordway, line 36: I think it's great that you handled the possibility of different inputs

lib/player.rb
```diff
@@ -0,0 +1,14 @@
...
+  def initialize(name, deck)
+    @name = name
+    @deck = deck
+  end
+  def has_lost?
+    if @deck.cards.empty? == true
+      true
+    else
+      false
+    end
```
jamisonordway, line 12: this works and is very easy to read. You could also consider something like
```
def has_lost?
  @deck.cards.empty? # this will evaluate to either true or false
end
```

lib/turn.rb
```diff
@@ -0,0 +1,123 @@
...
+      && (player1.deck.cards[2] == nil || player2.deck.cards[2] == nil)
+      @type = :war
+    elsif player1.deck.rank_of_card_at(0) == player2.deck.rank_of_card_at(0)\
+      && player1.deck.rank_of_card_at(2) == player2.deck.rank_of_card_at(2)
+      @type = :mutually_assured_destruction
+    elsif player1.deck.rank_of_card_at(0) == player2.deck.rank_of_card_at(0)
+      @type = :war
+    else
+      @type = :basic
+    end
```
jamisonordway, line 19: Firstly, I would not put this complex conditional branching in the `initialize` method, I would set up a helper method for `@type` like so:
```
def initialize(player1, player2)
  ...
  @type = determine_type
end

def determine_type
  if @player1.deck.rank_of_card_at(0) == @player2.deck.rank_of_card_at(0) # ...etc
end
```

Also, if you check for the `mutually_assured_destruction` condition **first**, then `war`, I think this conditional logic could be simplified.

spec/game_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+  describe '#start' do
+    it "successfully displays start message" do
+      game = Game.new
+      #not sure if this even works
+      expect { game.display_start_message }.to output(String).to_stdout
+    end
+    it "successfully starts game" do
+      game = Game.new
+      #cant figure out how to get the .start to work by inputting "go"
```
jamisonordway, line 34: You mentioned some overall difficulty with testing this class, which is fine-- our expectation in the rubric was for the code that does **not** require command line input/output to be fully tested.

spec/player_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      card3 = Card.new(:heart, 'Ace', 14)
+      deck = Deck.new([card1, card2, card3])
+      player = Player.new('Clarisa', deck)
+      expect(player.has_lost?).to eq(false)
+      player.deck.remove_card #2 cards remaining
+      player.deck.remove_card #1 card remaining
+      expect(player.has_lost?).to eq(false)
+      player.deck.remove_card #no cards remaining
+      expect(player.has_lost?).to eq(true)
+      expect(player.deck.cards).to eq([])
```
jamisonordway, line 39: nice job verifying that calling `remove_card` three times causes this change (`has_lost? => true` as well as `player.deck.cards` to be an empty array).

lib/turn.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  def war_winner
+    if player1.deck.rank_of_card_at(2) > player2.deck.rank_of_card_at(2)
+      @player1
+    else
+      @player2
+    end
+  end
+
+  def final_chance_war_winner
```
jamisonordway, line 52: When you add helper methods such as this one, `war_winner`, etc, keep in mind that these should be unit tested in isolation (for example, a test in `turn_spec.rb` which calls the `final_chance_war_winner` method directly and verifies its behavior).

PR #394, opened 2023-11-10
Turn class

DHowell1150:
This PR includes all of the Turn class test and specs for turns: 
- :basic.  When the top cards of the players' decks are not equal
- :war.  When the top cards of the players' decks are equal
- :mutually_assured_destruction. When the top cards are equal and the third cards of the players' decks are equal and there is no winner. 
 

INCLUDES METHODS:
- #winner. This indicates which player wins each turn if there is a winner.  
- #pile_of_cards.  Concats cards from players deck array to @spoils_of_war array. 
- #award_spoils(winner). Concats @spoils_of_war array into winner's deck array.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,12 @@
 ### Mentors
 This project is given with intent to work through with mentors at least twice through its duration.  The following are interactions I had with both Heidi and Jonathan:
 
-- 
+#### Heidi 11/6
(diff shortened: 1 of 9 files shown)


DHowell1150:
I just realized I haven't put in the feature to print "no winner" if a turn is mutually_assured_destruction."  I will need to put that in there.  Possibly in the #winner conditional.

